PR #11, opened 2019-06-10
Login route

SimeonSIL:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/11)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "react-dom": "^16.8.6",
     "react-localize-redux": "^3.5.3",
     "react-redux": "^7.0.3",
+    "react-router-dom": "^5.0.1",
     "react-scripts": "3.0.1",
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "@types/react": "16.8.19",
     "@types/react-dom": "16.8.4",
     "@types/react-redux": "^7.0.9",
+    "@types/react-router-dom": "^4.3.3",
     "@types/redux-thunk": "^2.1.0",
     "axios": "^0.19.0",
     "make-dir": "^3.0.0",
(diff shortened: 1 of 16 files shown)


SimeonSIL:
The purpose of this branch/pull request is redirecting users to and from the login page, and storing user jwt tokens in localStorage

jasonleenaylor:
___
*[src/components/Login/LoginComponent.tsx, line 77 at r2](https://reviewable.io/reviews/sillsdev/thecombine/11#-Lh61Oa0AgXIjcR-Node:-Lh61Oa0AgXIjcR-Nodf:b5gdab0) ([raw file](https://github.com/sillsdev/thecombine/blob/49c435ddadb6c38fb14194f9c1d1be2b43410b8c/src/components/Login/LoginComponent.tsx#L77)):*
> ```XML
>     if (user === "" || pass === "") {
>       // notify the user they need both a username and password
>       alert("Username and password cannot be blank");
> ```

There should be a material component that can be used for these type of error notifications. Alert is ok for now, but we should put this on a list of things to fix.

SimeonSIL:
___
*[src/components/Login/LoginActions.tsx, line 56 at r1](https://reviewable.io/reviews/sillsdev/thecombine/11#-Lh5kZMDEbIbQ41iD5bv:-Lh65khl4w06pYlfNG-K:b-d4cbhw) ([raw file](https://github.com/sillsdev/thecombine/blob/52d46df3eb266af69ecc193196b08b87ebb97eb0/src/components/Login/LoginActions.tsx#L56)):*
<details><summary><i>Previously, SimeonSIL wrote…</i></summary><blockquote>

I can't use the <Translate/> component within a browser alert. We can use material-ui dialogs, but I would need to add another component. Can I leave a TODO comment here?
</blockquote></details>

I got this alert to work, but it requires accessing the state within an action, which is not necessarily good practice. We'll look at using material-ui alerts later

PR #20, opened 2019-06-13
Linux deploy

jmgrady:
Initial package of Ansible playbooks for installing TheCombine on an Ubuntu 18.04 target.  All files are added to the "deploy" subdirectory of TheCombine.  TheCombine/deploy/README.md describes the usage of the setup script, "setup-target.sh" as well as the individual Ansible roles that are defined.

There are test.yml playbooks defined for the installation of node.js and the wifi access point.  Other test playbooks are stubs for the moment.

ansible-lint finds no issues.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/20)


jmgrady:
I would like to point out that this pull request includes adding a submodule from a repo that I created earlier on github and is itself forked from oblique/create_ap.  The changes I made were minor - mostly in how the configuration for the Access Point is built.
Having a submodule complicates things slightly but it is easy to trace the software to its origin.  I don't, however, expect to have further development on the create_ap repo that we will also want in TheCombine.
Please consider whether this should be kept as a submodule or if the files should be copied here and treated as  part of TheCombine.
Last of all, I had originally stated that we will want to use a wifi router instead of creating a wifi access point.  I have done some more experiments that suggest that a wifi access point may be viable so I have left this as part of the device setup.

jmgrady:
Comments on review comments:
Regarding, "The One True Editor", yes, I had seen a comment in LanguageForge about installing vim for Stephen.  That task name was for his benefit should he ever come across it. ;-) I have removed it from the playbook (partly because the test for Server instead of Desktop was ugly) and will put in in a personal playbook  (cf. "Effective STL", Scott Meyers, Item 5)

Regarding the grub defaults, I changed the sense of the comment

Regarding the wifi configuration, I removed the creation of create_ap.hosts.  This doesn't work until we get the virtual hosts setup for apache.  Sorry, I jumped the gun here.

PR #60, opened 2019-06-20
Upgrade goal selector scroll ui

Micah-Schiewe:
In addition to centering ui/fixing its style:
* Adds a custom right-click drop-down menu for goal cards
* Repairs jerry-rigged unit tests for GoalSelectorScroll

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/60)

### DIFF
--- a/src/components/ContextMenu/ContextMenu.tsx
+++ b/src/components/ContextMenu/ContextMenu.tsx
@@ -1,0 +1,89 @@
+import { Menu, MenuItem } from "@material-ui/core";
+import React, { ReactElement } from "react";
+import {
+  LocalizeContextProps,
+  withLocalize,
+  Translate
+} from "react-localize-redux";
+
+export type MenuType = [string, () => void];
+
+export interface ContextMenuProps {
+  options: MenuType[];
+  anchorName: string;
+}
+export interface ContextMenuState {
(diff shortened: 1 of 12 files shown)


Micah-Schiewe:
___
*[src/components/GoalTimeline/GoalSwitcher/GoalSelectorScroll/GoalSelectorScroll.tsx, line 262 at r1](https://reviewable.io/reviews/sillsdev/thecombine/60#-LhpnJ5s9pioRrOWBaO3:-Li8ZxEP6hYmRRPPDpD-:b2uvjbx) ([raw file](https://github.com/sillsdev/thecombine/blob/19dc8fba92af2ef43bd3d816386f8ae80b4b7aca/src/components/GoalTimeline/GoalSwitcher/GoalSelectorScroll/GoalSelectorScroll.tsx#L262)):*
<details><summary><i>Previously, tyasw wrote…</i></summary><blockquote>

When I find myself needing to add section comments, I realize the class is too big. I would try to break this class up into smaller ones. Maybe for another pull request, though. This component is only referenced in the GoalSwitcher component, so you should be able to do this anytime.
</blockquote></details>

Definitely not in this commit (we need to get it added in), but I'll see if there's any breakdown I can do.

PR #121, opened 2019-07-02
Add transient for IUserEditRepository

sam-skinner:


Backend/Startup.cs
```diff
@@ -98,6 +98,7 @@ public void ConfigureServices(IServiceCollection services)
             // User role types
             services.AddTransient<IUserEditContext, UserEditContext>();
             services.AddTransient<IUserEditService, UserEditService>();
+            services.AddTransient<IUserEditRepository, UserEditRepository>();
```
tyasw: Good catch.

PR #151, opened 2019-07-09
Fix merge words (Fix #143)

### LINKED ISSUE #143: Split word during merge
I want to be able to tag senses during a merge that should be split off into their own word for further processing later. All senses with their state set to seperate should be put in their own new word.

I have a graph demonstrating the use case for what I am asking for
image

sam-skinner:
(Fix #143)

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/151)


SamDelaney:
resolve #143

PR #189, opened 2019-07-19
Further duplicate finder optimization

SamDelaney:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/189)


SamDelaney:
The remaining test that does not pass should be fixed by #188

PR #217, opened 2019-07-30
Add Pronunciations component

jasonleenaylor:
* Always has a record icon
* Has one play icon for each audio file

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/217)


Micah-Schiewe:
___
*[src/components/Pronunciations/index.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/thecombine/217#-Ll3d-gU9daj_Vi1W_yE:-Ll3eA-490VhibxbCVrs:b-pumkrt) ([raw file](https://github.com/sillsdev/thecombine/blob/3f6eef7c4248bc24b80a0f1032b098727bfb7a45/src/components/Pronunciations/index.ts#L7)):*
<details><summary><i>Previously, Micah-Schiewe wrote…</i></summary><blockquote>

> ```
> translate: getTranslate(state.localize),
> ```
From what I've seen so far, there's usually a better way of accessing translations than having to link to the translate function. Usually, you can use <Translate id="address" /> in the render to show translated text.
</blockquote></details>

Okay, there WAS a code snippet in there, but Reviewable decided to remove it. Here's an alternative to linking translate in the props:
`<Translate id="address to translations in translations.json" />`

jasonleenaylor:
___
*[Backend/Controllers/AudioController.cs, line 75 at r3](https://reviewable.io/reviews/sillsdev/thecombine/217#-Llhqz4R0EBEn7zChS-_:-LlliTWE5zK1_iivOPkV:b-dftsgq) ([raw file](https://github.com/sillsdev/thecombine/blob/5bf43109776e74a0d321e8bfb9cc0fff9c944a69/Backend/Controllers/AudioController.cs#L75)):*
<details><summary><i>Previously, SimeonSIL (Simeon) wrote…</i></summary><blockquote>

This might not be an issue, but if we upload multiple files to the same word, and they don't have filenames on them, they'll conflict
</blockquote></details>

true. I should probably return an error here instead of trying to be forgiving.

PR #232, opened 2019-07-31
Fix loading no items in ReviewEntries

Micah-Schiewe:
* Fix some React key issues
* Index is still used for some items, but since re-arranging senses aren't allowed, this isn't much of an issue

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/232)

### DIFF
--- a/src/goals/ReviewEntries/ReviewEntriesComponent/CellComponents/AlignedList.tsx
+++ b/src/goals/ReviewEntries/ReviewEntriesComponent/CellComponents/AlignedList.tsx
@@ -5,6 +5,7 @@ import { Add } from "@material-ui/icons";
 export const SPACER: string = "spacer";
 interface AlignedListProps {
   contents: ReactNode[];
+  listId: string;
   bottomCell: ReactNode | typeof SPACER | null;
 }
 
@@ -16,6 +17,7 @@ export default class AlignedList extends React.Component<AlignedListProps> {
           <Grid
             item
             xs
+            key={`alignedList:${this.props.listId}:${index}`}
             style={
               this.props.bottomCell || index + 1 !== this.props.contents.length
                 ? {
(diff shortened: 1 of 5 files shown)


Micah-Schiewe:
___
*[src/goals/ReviewEntries/ReviewEntriesComponent/CellComponents/DomainCell.tsx, line 91 at r1](https://reviewable.io/reviews/sillsdev/thecombine/232#-LlCQPoB84TyPRCi1WYQ:-LlCRgfp5lbXT1BSeCc9:b2geofb) ([raw file](https://github.com/sillsdev/thecombine/blob/64f8204186b088a8201211e313ed3b6e3249ab7c/src/goals/ReviewEntries/ReviewEntriesComponent/CellComponents/DomainCell.tsx#L91)):*
<details><summary><i>Previously, tyasw wrote…</i></summary><blockquote>

Just wondering, why are there two colons here (::)?
</blockquote></details>

I think I was just creating some visual difference between it and the colons between numbers in the key. It's not an official convention, if that's what you're wondering.

Micah-Schiewe:
___
*[src/goals/ReviewEntries/ReviewEntriesComponent/CellComponents/DomainCell.tsx, line 91 at r1](https://reviewable.io/reviews/sillsdev/thecombine/232#-LlCQPoB84TyPRCi1WYQ:-LlCT4Qy8JZBCntcs9Ac:bns0532) ([raw file](https://github.com/sillsdev/thecombine/blob/64f8204186b088a8201211e313ed3b6e3249ab7c/src/goals/ReviewEntries/ReviewEntriesComponent/CellComponents/DomainCell.tsx#L91)):*
<details><summary><i>Previously, SimeonSIL (Simeon) wrote…</i></summary><blockquote>

As a side note, if you're going to use the index as the key, just use the index `key={index}`. Keys are only used to distinguish between children in the same list, so React won't be confused between item 1 of domains and item 1 of senses.
</blockquote></details>

Okay, gotcha. In that case, a lot of this can be simplified (I assumed keys were slightly more global than that)

PR #235, opened 2019-08-01
Audioavatar controller

SILMarkFuller:
@jasonleenaylor should merge this into his branch and test it, I'm not sure how I'm streaming the audio is what you're expecting.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/235)


jasonleenaylor:
___
*[Backend/Services/WordApiServices.cs, line 233 at r2](https://reviewable.io/reviews/sillsdev/thecombine/235#-LlCk_yM08uW6oQcWlx5:-LlCk_yM08uW6oQcWlx6:bw6yt13) ([raw file](https://github.com/sillsdev/thecombine/blob/bbb88568b05209276831c01906eaba148a73e1d1/Backend/Services/WordApiServices.cs#L233)):*
> ```Smalltalk
>                 return null;
>             }
>             filepath = Path.Combine(filepath, listOfDirs.Single(), wordId, ".mp3"); //there should only be one dir in that file
> ```

Hmm. mp3 probably isn't always the right answer.

jasonleenaylor:
___
*[Backend/Services/WordApiServices.cs, line 233 at r2](https://reviewable.io/reviews/sillsdev/thecombine/235#-LlCk_yM08uW6oQcWlx5:-LlCko1C2Q4-_Vn-utmq:b-myuc9l) ([raw file](https://github.com/sillsdev/thecombine/blob/bbb88568b05209276831c01906eaba148a73e1d1/Backend/Services/WordApiServices.cs#L233)):*

Hmm. mp3 probably isn't always the right answer.
</blockquote></details>

We'll sort that out later. Also we need to support multiple audio files.

PR #236, opened 2019-08-01
Net config

jmgrady:
Adds the following functionality to the deploy scripts:

 - Add a self-signed CA certificate, CombineCA.pem, to sign certificate for Kestrel backend server
 - Generate a certificate signed by CombineCA.pem for Kestrel backend server and trusts the certificate
 - Add certificate passwords to an Ansible vault file for development
 - Change Apache configuration file as follows:
    - http connection to port 80 is redirected to https on port 443
    - Reverse Proxy configuration corrected
 - Create `combine` user  on the target
 - Install Backend server as a service and run as user `combine`
 - Update setup-target.sh script to add a `-b` option which will:
    - change the baseURL in src/backend/index.tsx from `localhost:5001` to `thewordcombine.org`
    - build the react UI
    - build the Backend service
 - Update vagrant image to have .NET SDK 2.1 to match project requirements (version 1.0.4)
 - Generate `appsettings.Production.json` with configuration for Kestrel server and configurable logging levels (specified in Ansible variables)
 - Setup Backend server to UseForwardedHeaders
 - Configure WiFi access point to set the local domain to thewordcombine.org

In addition,
 - Add submodule for common ansible roles (`sillsdev/ops-ansible-common-roles` is added as `roles_common`) The ssl_config role is used to generate the certificate for the Apache web server.
- mkcombine script is now updated as part of provisioning the vagrant vm; no need to create a new image.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/236)


deploy/roles/the_combine_app/files/CombineCA/CombineCA.key
```diff
@@ -0,0 +1,30 @@
+-----BEGIN RSA PRIVATE KEY-----
```
jasonleenaylor: So, what are the consequences of this key being public? When we get our real cert we won't want it in source control...

jasonleenaylor:
___
*[deploy/roles/the_combine_app/files/CombineCA/CombineCA.key, line 1 at r5](https://reviewable.io/reviews/sillsdev/thecombine/236#-LlCZ_o_lxWDxHweQGT4-r5-1:-LlIiSofDenopkPtRk_D:b-24bx6g) ([raw file](https://github.com/sillsdev/thecombine/blob/7dd5c1977bd5c419b24aead69f48c4189b43289a/deploy/roles/the_combine_app/files/CombineCA/CombineCA.key#L1)):*

My expectation is that this key and cert will go away when we get our real cert (or is that only for production machines?)  If we are not concerned about the development self-signed cert being in source control then I think this is o.k.  Is there  shared location outside of github where we can post CA certs for development use?
</blockquote></details>

I guess this is fine for dev, we don't have a good space set-up for this. We just need to make sure and be extra careful to keep the real certs out of the repository.

PR #293, opened 2020-04-25
Backend Cleanup

johnthagen:
Style guide followed taken from the official MS C# style guide:

- https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/identifier-names
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions

Summary of changes:

- No functionality is changed
- Fix issues that do not conform to MS C# conventions (essentially get the project to all be following the same style guide, in this case the standard C# style guide)
- Rename `BackendFramework.ValueModels` namespace to match C# guidelines
- Rename `BackendFramework.Contexts` namespace to match C# guidelines
- Fix spelling mistakes
- Make variables `const` where possible
- Make methods `static` where possible
- Sort `using` directives
______

When merging, squash commits.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/293)

### DIFF
--- a/Backend.Tests/AudioControllerTests.cs
+++ b/Backend.Tests/AudioControllerTests.cs
@@ -34,7 +34,7 @@ namespace Backend.Tests
             _audioController = new AudioController(_wordrepo, _wordService, _permissionService);
 
             var util = new Utilities();
-            Directory.Delete(util.GenerateFilePath(Utilities.Filetype.dir, true, "", ""), true);
+            Directory.Delete(util.GenerateFilePath(Utilities.FileType.Dir, true, "", ""), true);
         }
 
         private static string RandomString(int length = 16)
--- a/Backend.Tests/AudioControllerTests.cs
+++ b/Backend.Tests/AudioControllerTests.cs
@@ -9,6 +9,7 @@ using Microsoft.AspNetCore.Mvc;
 using NUnit.Framework;
 using System;
 using System.IO;
+using BackendFramework.Models;
 
 namespace Backend.Tests
 {
--- a/Backend.Tests/AudioControllerTests.cs
+++ b/Backend.Tests/AudioControllerTests.cs
@@ -53,18 +53,18 @@ namespace Backend.Tests
         public void TestAudioImport()
         {
             // Get path to sound in Assets folder, from debugging folder.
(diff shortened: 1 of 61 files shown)


johnthagen:
I tested the following to ensure it works on with these changes:

- Register new user
- Login
- Add avatar picture
- Create a new project
- Add a new entry (vernacular, glosses, and audio clip)
- Reviewed entry / listened to audio in Review Entries

jmgrady:
Hi John,

Sorry that I did not see your PR earlier.  I know that Jason is busy but he is the best one to review any C# changes.  I don't have the experience yet to be qualified to review the changes.  I will review it - mostly for my education.  If I see anything,  I certainly let tippy know. 

⁣Jim

"In practice, the difference between theory and practice is greater than it is in theory."​

johnthagen:
@jmgrady Sounds good. Since there aren't any differences in operation, I was hoping you could run through and test it out on your machine.

johnthagen:
@jasonleenaylor Thanks for the review.

TLDR; I too agree with using type inference (`var`) everywhere, and I will update the PR to reflect this.

________

### Longer Explanation

Since I didn't know what SIL's coding guidelines were, I read up on the Microsoft guidelines (referenced in the first message) and applied them in lieu of my personal preferences.

According to those guidelines (https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#implicitly-typed-local-variables):

> Use implicit typing for local variables when the type of the variable is obvious from the right side of the assignment, or when the precise type is not important.

> Do not use var when the type is not apparent from the right side of the assignment.
C#

```c#
// When the type of a variable is not clear from the context, use an
// explicit type.
int var4 = ExampleClass.ResultSoFar();
```

So I applied that rule and only used `var` when the right side was obvious (the most common case was a constructor, but also `.toString()` calls, etc.)

I found this was very challenging to keep consistent and I agree with you that letting the compiler and IDE infer the type leads to a much cleaner experience and easier refactoring.

_______

### Moving Forward

Would you concur if we adopted the MS C# guidelines with an exception that `var` be used wherever possible?

Issue #300, opened 2020-05-01
Add Site Admin role


jasonleenaylor:
Currently there is no way to delete users or projects.
We need to add a new role for site administrators and then add endpoints for deleting users and projects which are restricted to that role.
Then we need to add a Site Settings page available to site administrators which will provide the interface to access the delete project and delete user endpoints.

johnthagen:
This role currently exists, but is simply unused:

https://github.com/sillsdev/TheCombine/blob/177fea2ab7fdc23d4f15c5de702dc6cd464745d7/Backend/Models/UserRole.cs#L90

Will use that for the basis of the new endpoints.

johnthagen:
@jasonleenaylor 

> We need to add a new role for site administrators and then add endpoints for deleting users and projects which are restricted to that role.

Should Project Admins be able to delete projects they create?

https://github.com/sillsdev/TheCombine/blob/177fea2ab7fdc23d4f15c5de702dc6cd464745d7/Backend/Models/UserRole.cs#L92-L93

johnthagen:
Actually, using `Permission.DatabaseAdmin` is a little strange since `UserRole.Permissions` is designed for permissions a user has towards a *project*. `User.IsAdmin` seems more appropriate:

https://github.com/sillsdev/TheCombine/blob/177fea2ab7fdc23d4f15c5de702dc6cd464745d7/Backend/Models/User.cs#L57

johnthagen:
Many of these endpoints exist, but are wrapped in `#if DEBUG` blocks, presumably because this functionality hadn't been completed.

PR #303, opened 2020-05-01
Site Admin Endpoints

### LINKED ISSUE #297: Refactor "magic values" in Controller IsProjectAuthorized Calls
There any many calls similar to in `Controllers`:

- https://github.com/sillsdev/TheCombine/blob/8f56dd4adbe05c025916a1f601a53db81a5b05d1/Backend/Controllers/FrontierController.cs#L32
- https://github.com/sillsdev/TheCombine/blob/8f56dd4adbe05c025916a1f601a53db81a5b05d1/Backend/Controllers/LiftController.cs#L44

But it's not clear what these magic strings are. Ideally they should be replaced with an `enum`, if possible.

### LINKED ISSUE #306: seedAdmin.bat script does not store password correctly
After running `seedAdmin.bat` not that it puts the default admin password in as a string, rather than hashing it, which results in it not being able to be used when logging into the app.

image

Relates to #305

johnthagen:
Backend portion for #300
Closes #297 
Closes #306 

- Enable `User` DELETE endpoints
- Enable `Project` DELETE all endpoint
- Create `setAdminUser.bat` script to set users as database admins

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/303)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,18 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
-# dependencies
+# Dependencies
 /node_modules
 /.pnp
 .pnp.js
 
-# testing
+# Testing
 /coverage
 /Backend.Tests/coverage.json
 
-# production
+# Production
 /build
 
-# misc
+# Misc
 .DS_Store
 .env.local
 .env.development.local
@@ -27,7 +27,7 @@ yarn-error.log*
 # Visual Studio
 .vs/*
 
-#Backend
+# Backend
 Backend/.vs/**/*
 Backend/bin
 Backend/obj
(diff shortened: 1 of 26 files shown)


jasonleenaylor:
___
*[Backend/Services/PermissionService.cs, line 52 at r1](https://reviewable.io/reviews/sillsdev/thecombine/303#-M6Xccld1dkZUDqouFFU:-M6Xccld1dkZUDqouFFV:b4woros) ([raw file](https://github.com/sillsdev/thecombine/blob/177fea2ab7fdc23d4f15c5de702dc6cd464745d7/Backend/Services/PermissionService.cs#L52)):*
> ```Smalltalk
>         }
> 
>         public bool IsProjectAuthorized(string value, HttpContext request)
> ```

FYI, we ran into some trouble with getting the Enum to round trip successfully so we fell back to this in order to get something shipped. So watch for trouble with that. I was pretty sure that there should be a solution but our first attempts didn't work.

johnthagen:
> FYI, we ran into some trouble with getting the Enum to round trip successfully so we fell back to this in order to get something shipped. So watch for trouble with that. I was pretty sure that there should be a solution but our first attempts didn't work.

Understood, will look out for this. For now, the changes are only internal to the backend (everything is still serialized as `int`s). I was just trying to make the internal API's more type-safe.

PR #310, opened 2020-05-09
Frontend Implementation of Site Admin

### LINKED ISSUE #300: Add Site Admin role

### LINKED ISSUE #308: Database Admin Users do not see all projects
When a Database Admin (`User.isAdmin = true`) logs in, they should be able to see all `Project`s.

johnthagen:
Closes #300

Closes #308

- Don't automatically open a browser tab when backend is launched. This makes restarting the backend during development more pleasant.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/310)


johnthagen:
@jasonleenaylor If someone else wants to take this over too, that is totally fine by me. At the time I was the only person working on the backend and frontend, but now that we have a frontend team, it probably makes sense to pass this off. The backend side of things *should* be ready now.

johnthagen:
@imnasnainaec Just an FYI, I'm currently trying to work out a way to create new admin users directly in #413, but for now you should be able to test with https://github.com/sillsdev/TheCombine#create-database-admin-user

imnasnainaec:
The project list and user list components within the SiteSetttings page still need to be developed. This pr is primarily to get the underlying page in place.

imnasnainaec:
Resolves #308

PR #313, opened 2020-05-09
Remove create ap submodule

jmgrady:
The create_ap code was originally a submodule that points to a repo outside of SIL's control.  The original repo was forked from https://github.com/oblique/create_ap so that updates could be pulled from the original source.

In order to remove links to repos outside of sillsdev this PR replaces the submodule with a copy of the code that is used to create the access point on the NUC.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/313)


johnthagen:
To me this seems like a good change as we are no longer dependent on a fork in @jmgrady's GitHub.

Issue #301, opened 2020-05-01
Fix LIFT roundtrip with FLEx


jasonleenaylor:
An important use case is to:

1. Export to LIFT from FLEx, zip up the folder, and import it into The combine
2. Clean up the data by deleting some entries, merging others, and or adding new entries through Data Entry
3. Export Project from The Combine (in the Project Settings screen)
4. Import that LIFT data into FLEx and see the changes occur

I have encountered some crashes in Export Project trying to verify this workflow.

PR #340, opened 2020-05-28
Fix lift export with audio

Rapdorian:
Fixes import/export and uploading audio so that they all use the same location for storing audio.

This should solve #301 that said I don't have FLEx on my machine to test with. However I have been able to successfully roundtrip projects through The Combine with and without audio

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/340)


johnthagen:
@Rapdorian I merged master into your branch to address a new CI build step we are running for the backend. This shouldn't affect your code, but wanted to give you a heads up to pull your branch down again locally.

Issue #299, opened 2020-04-26
Backend compile warnings


johnthagen:
```
> dotnet test Backend.Tests/Backend.Tests.csproj
Controllers\AudioController.cs(38,42): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [C:\Users\User\GitHub\TheCombine\Backend\BackendFramework.csproj]
Controllers\ProjectController.cs(282,42): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [C:\Users\User\GitHub\TheCombine\Backend\BackendFramework.csproj]
```

johnthagen:
We could add

```
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```

to treat warnings as errors and fail CI if warnings are present.

PR #342, opened 2020-05-29
Treat backend warnings as errors

johnthagen:
- Treat all warnings as errors so that warnings cannot slip into `master`
- Suppress two `async` related warnings locally

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/342)


Backend/Controllers/AudioController.cs
```diff
@@ -34,7 +34,9 @@ public AudioController(IWordRepository repo, IWordService wordService, IPermissi
         /// <returns> Audio file stream </returns>
         [AllowAnonymous]
         [HttpGet("{wordId}/download/audio/{fileName}")]
+#pragma warning disable 1998
```
jasonleenaylor, line 37: When we do this I prefer to also have a comment that explains why the compiler was wrong

johnthagen: @jasonleenaylor Part of what I wanted to discuss is if there was a way to `await` some of the operations in these functions. I'll do some digging today to see if either of these functions can be made properly `async`.

johnthagen: @jasonleenaylor Comments have been added.

imnasnainaec:
This resolves #299

Issue #366, opened 2020-05-30
"Forgot Password" doesn't do anything in Login Page


shyung18:

johnthagen:
Priority for this is higher since it is blocking #412 now.

PR #368, opened 2020-05-30
Cleanup duplicated password hashing 

### LINKED ISSUE #367: Consolidate duplicated password hashing functionality
Fix https://github.com/sillsdev/TheCombine/issues/344#issuecomment-636326644

johnthagen:
Closes #367

- Consolidates and de-duplicates password hashing logic
- Change from SHA-1 algorithm to recommended SHA-256 to increase security
- Increase password iteration count from 10,000 to 216,000 to match current security best practices in 2020

Note: As a result of this change, users in current databases will need to be dropped and re-added as the password hashing implementation has been made more secure.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/368)


johnthagen:
We are deferring merging this until #366 is implemented, as this will require all users to be re-created in the production deployment.

johnthagen:
@jasonleenaylor I've reverted the breaking changes and left the rest of the refactor so that the refactor/cleanup can be merged now. I'll open a separate issue for strengthening password security.

PR #374, opened 2020-06-01
Switch project

### LINKED ISSUE #333: Add list of other projects that a user is on to the project settings screen
Right now the only way to switch projects or even see projects is a screen when you first login.

Add a new component that lists projects to the project settings screen and allow the user to choose another project and switch to it.

shyung18:
Closes #333 

Able to switch between projects in the project settings view. 
It gives a list of projects and highlights the current project.

Problem: Does not update the project name.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/374)


StevesBro:
___
*[src/backend/index.tsx, line 200 at r3](https://reviewable.io/reviews/sillsdev/thecombine/374#-M8l8TQL6JSMWxJQb1Lr:-M8l8TQL6JSMWxJQb1Ls:bljti1g) ([raw file](https://github.com/sillsdev/thecombine/blob/b7f9995cb361addfa96c320469f5fb2e75702ce6/src/backend/index.tsx#L200)):*
> ```XML
> 
> export async function getAllProjectsByUser(user: User): Promise<Project[]> {
>   let projectIds: string[] = Object.keys(user.workedProjects);
> ```

A project isn't added to a user's workedProjects until the user changes something in the project.
If a user creates a project and logs out without changing anything, then that project won't show in the project list.

Issue #339, opened 2020-05-28
Going to a specific semantic domain does not always work


stevendyk:
Steps to reproduce:
1. In "Find a Domain" type 8.3.1.2.3 and press Enter
2. Notice nothing happened
If you type 8.3.1.2 you can see see that there is an 8.3.1.2.3

Expected behavior:
The tree navigates to 8.4.1.2.3.

Hypothesis:  The find is not working for more than 4 digits

![TopOfTree](https://user-images.githubusercontent.com/25255423/83155412-d2deb180-a0c6-11ea-8397-c722f2069bee.png)

Issue #320, opened 2020-05-22
Upgrade frontend dependenices - tests break on newest react-scripts


johnthagen:
# Important

When this issue is resolved, be sure to **remove** the following from `package.json`:

```json
    "rules": {
      "no-undef": "off"
    }
```

This was added in #319 in order to allow ESLint to pass and put on CI without blocking that on this issue.

# Problem

The frontend dependencies are locked at pretty old versions. 

To update:

```bash
$ npm i -g npm-check-updates
$ ncu -i
$ npm install
```

The issue is that after upgrading, it looks like some libraries have changed enough to break the tests:

```
Snapshot Summary
 › 9 snapshots failed from 5 test suites. Inspect your code changes or press `u` to update them.

Test Suites: 10 failed, 57 passed, 67 total
Tests:       32 failed, 200 passed, 232 total
Snapshots:   9 failed, 9 total
Time:        35.085s
Ran all test suites.
```

The frontend does still build, however.

johnthagen:
Tests start failing moving `react-scripts` from `3.0.1` to `3.1.2` (latest is `3.4.1`).

```
Test Suites: 5 failed, 62 passed, 67 total
Tests:       23 failed, 209 passed, 232 total
Snapshots:   9 passed, 9 total
Time:        33.833s
Ran all test suites.
```

Steps to resolve:

1. Get the current test suite passing without warning (there are currently a lot of warnings printed out even as is)
2. Bump `react-scripts` up one version, see what is failing and research what changed from 3.0 to 3.1 that needs to be fixed (changes to Jest, etc.).
3. Repeat step 2 for 3.1 -> 3.2, 3.2 -> 3.3, 3.3 -> 3.4.

PR #392, opened 2020-06-04
Added unit tests for issue #339

charjhall:
Modifies MockSemanticTree to increase the depth of the tree and adds two unit tests for issue #339

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/392)

### DIFF
--- a/src/components/TreeView/tests/TreeViewHeader.test.tsx
+++ b/src/components/TreeView/tests/TreeViewHeader.test.tsx
@@ -20,30 +20,9 @@ var treeMaster: ReactTestRenderer;
 var treeHandle: TreeViewHeader;
 const MOCK_ANIMATE = jest.fn();
 
-// Mock out window
-const ADD_LISTENER = window.addEventListener.prototype;
-const REMOVE_LISTENER = window.removeEventListener.prototype;
-
-// TODO: Should this lint be disabled?
-// eslint-disable-next-line no-native-reassign
-window = {
-  ...window,
-  addEventListener: jest.fn(),
-  removeEventListener: jest.fn(),
-};
-
 beforeEach(() => {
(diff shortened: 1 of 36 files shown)


src/components/TreeView/tests/TreeViewHeader.test.tsx
```diff
@@ -74,6 +74,33 @@ describe("Tests TreeViewHeader", () => {
     expect(event.target.value).toEqual(TEST);
   });
 
+  it("Search & select domain switches on a length 5 number", () => {
+    var leafnode: SemanticDomainWithSubdomains =
```
johnthagen, line 78: Can this be changed to `const leafnode` to make the binding immutable?

charjhall: Yes most certainly. I was actually unsure of whether to declare this as a const, var, or let. Is it good practice to declare const if it is not changing?

johnthagen: Yes, in modern JS/TS you should default to always using `const` (though this code base could be cleaned up a bit in that regard). This makes the binding immutable and can avoid subtle errors of variables being updated or reassigned unintentionally.

If the value needs to be changed, use `let`.

`var` should be avoided in most cases.

johnthagen: Also, JS/TS conventions are to use `lowerCamelCase`, so style-wise I think the best result is:

```typescript
const leafNode: ...
```

johnthagen:
@charjhall Just an FYI, due to #320 we are on a somewhat older version of Jest and such. So if you run into things you see online that are different to how they can be done here, that might be part of the reason.

Issue #344, opened 2020-05-29
Create initial database admin on target systems


jmgrady:
The playbooks that install the combine on the demo server and the NUCs need to create an initial, known database user with admin privileges.  Requirements:
1. When TheCombine is first started, there shall be a user, `admin`, pre-installed in the database;
2. The `admin` user shall have the same password across all systems;
3. The `admin` password shall be encrypted in the github repo (e.g. stored in an Ansible vault variable or file)

johnthagen:
Relates to #305

jmgrady:
John (@johnthagen), Jason (@jasonleenaylor),
I would like your thoughts on this topic.  I have thought of two methods to solve this problem:
1. Create an executable that can be run on the target system that creates/updates an admin user.
   In order to do things the ansible way, I would want the executable to return an exit code that indicates "No change", "Successful change", or "Failure".
2. Have the backend create the user at first start (it already creates the database, I believe).
   In this case, we could create environment variables for the backend process that tells it what username/password to use.  I don't think that we will want to worry about multiple users being created or for non-admin users to be created.  If I understand correctly, the main purpose here is to have an admin user defined who can be used to make other users admins if needed.

I have not delved into the password hashing code very deeply but if it is equivalent to one of the methods described [here](https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#how-do-i-generate-encrypted-passwords-for-the-user-module) then the ansible playbook can generate the password field and use the mongo shell to create the user as needed.

I'm sure that there are other ways to solve this as well.  Let me know your ideas.

johnthagen:
The password hashing algorithm is defined here:

https://github.com/sillsdev/TheCombine/blob/b3cfd3623c158036f2d73bb9a133b76baafdd44e/Backend/Services/UserApiServices.cs#L34-L66

And duplicated here:

https://github.com/sillsdev/TheCombine/blob/b3cfd3623c158036f2d73bb9a133b76baafdd44e/Backend/Services/UserApiServices.cs#L158-L190

At a minimum this logic needs to be consolidated in the backend, so I'll work on that.

@jasonleenaylor How have you handled this in the past on other .NET Core applications?

The frameworks I'm most familiar (Python/Django) provides a [`createsuperuser`](https://docs.djangoproject.com/en/3.0/ref/django-admin/#createsuperuser) command line management command for this.

jmgrady:
@johnthagen 
[...]

PR #413, opened 2020-06-06
Create Admin User CLI

### LINKED ISSUE #411: Create Admin User Script
Create a way to programmatically create admin users in the backend from the command line.

https://github.com/sillsdev/TheCombine/issues/344#issuecomment-636158197

johnthagen:
Closes #411

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/413)


johnthagen:
A few things make this implementation difficult:

- ASP.NET Core doesn't have a [`createsuperuser`](https://docs.djangoproject.com/en/3.0/ref/django-admin/#createsuperuser) equivalent (that I know of) that can create a user into the database independent of starting up the main application
- We can't allow Docker to start up the backend container normally, and then try to add the user, because the second instance of our application (and it's Kestrel) that tries to start will try to bind to the same port and fail.
- It's somewhat tricky to get exit codes to propagate our of Docker compose (but not impossible). It just means we can't launch with `--detach`

johnthagen:
@jmgrady This is ready for initial review. It *works*, albeit a little more involved than I would have liked. I might try to optimize/simplify it some more but wanted to get an initial review from you first.

johnthagen:
___
*[README.md, line 156 at r5](https://reviewable.io/reviews/sillsdev/thecombine/413#-MALOsl05zZZW0FTJH0c:-MAlTBXp53MOVJ67_lv-:b65lo90) ([raw file](https://github.com/sillsdev/thecombine/blob/8891cb1d9fc29a09b3f21d5f135fba33284c6f9b/README.md#L156)):*

> ```
> (Development Only)
> ```
Technically, if we have the source code on the targets, this is not just development, correct?  (Assuming one could ssh to the target and run `docker exec ...`
</blockquote></details>

We probably won't have Node/`npm` on the host, so these instructions won't work on a server install target, and aren't designed for production use.

johnthagen:
___
*[.env.web.auth.template, line 2 at r10](https://reviewable.io/reviews/sillsdev/thecombine/413#-MBnxWr-2nL9Z77krpau:-MBsvd0Y4dUOJOW-EYQ4:bb4k2eb) ([raw file](https://github.com/sillsdev/thecombine/blob/dce7ef7f55bc7307a3c144e42cf7e0f1ebd08f76/.env.web.auth.template#L2)):*

> ```
> ASPNETCORE_ADMIN_USERNAME=
> ASPNETCORE_ADMIN_PASSWORD=
> ```
I think that these environment variable names are misleading.  These are not admin username & password for ASP .NET Core, they are admin username & password for the Backend or the COMBINE.  ASPNETCORE is an implementation artifact of the current backend.
</blockquote></details>

When I started working on Combine there were some conventions in place for the environment variables the backend were using, they were all prefixed with `ASPNETCORE_` so I assumed there was some common convention they were following.

So I have been sticking to that convention for consistency thus far.

I agree in the larger context though, it would be better to have app-specific names for these. Since James is working on another PR that is using these names I think it would be cleanest to wait for that PR and this PR to be merged and then have a PR that just cleans up the names.

I opened a separate issue for this: https://github.com/sillsdev/TheCombine/issues/501

johnthagen:
___
*[README.md, line 153 at r5](https://reviewable.io/reviews/sillsdev/thecombine/413#-MALMD1vD8Patph5VAOF:-MBsyd7aEXN6PKtAsARB:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/8891cb1d9fc29a09b3f21d5f135fba33284c6f9b/README.md#L153)):*

> ```
> **Important**": Remove the `ASPNETCORE_*` environment variables from
> `.env.web` or subsequent launches will fail because the backend will try
> to re-create the same user name again.
> ```
As I commented in issue #344, an executable to create an admin user should have three return states:
1. Success: User created/updated (if the user exists but the password is different, then update the password);
2. Success: No change required
3. Failure: something went wrong.
This was assuming a separate executable that would be called as part of installing the system, not part of calling the backend but much of the thinking applies.

The reason for updating an existing user is that it makes administering multiple targets more manageable.  If we decide that we need to change the admin password, we can use the scripts to do it - not have to login to TheCombine on each target and change the password.  Perhaps changing the password is better done as a separate script because we probably don't want to be setting the new password without verifying that the person running the script knows the old password.  Also, having the docker file differ between the first run and all subsequent runs is not workable for production.

If you made the script return success when the user exists already and the password matches, then that at least gets rid of the need to change the Docker file.  This makes sense if you think of the job of `create-admin-username` in terms of the desired end-state rather than making a state transition.
</blockquote></details>

Done.

johnthagen:
___
*[Backend/launch.sh, line 4 at r5](https://reviewable.io/reviews/sillsdev/thecombine/413#-MALPU3kD4vybJrzqtLu:-MBsyeTA0AflIjsQdZlP:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/8891cb1d9fc29a09b3f21d5f135fba33284c6f9b/Backend/launch.sh#L4)):*

> ```
> #!/bin/sh
> 
> if [ -n "$ASPNETCORE_ADMIN_USERNAME" ]; then
>     dotnet BackendFramework.dll create-admin-username=$ASPNETCORE_ADMIN_USERNAME
> else
>     dotnet BackendFramework.dll
> fi
> ```
Why not put this logic in the Backend itself?  (assuming that it is not an error if the user exists with the supplied password)  When we discussed a command-line method for creating the user, I was expecting that there was a separate executable with perhaps shared source code.  This is _much better_ in that you don't have to worry about both executables having the same ASPNETCORE_JWT_SECRET_KEY.

Also, regarding my comments on the README.md, if the user exists and the password does not match, then perhaps a warning could be logged.  I'm not sure of the best approach.  Ideally, I would like it to be part of building the container rather than starting it up.  Do you have any ideas for this?
</blockquote></details>

Done.

johnthagen:
___
*[Backend/Startup.cs, line 184 at r5](https://reviewable.io/reviews/sillsdev/thecombine/413#-MALSV3aBVFSlYZgZ00M:-MBsyfQ30mcysOsoywXV:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/8891cb1d9fc29a09b3f21d5f135fba33284c6f9b/Backend/Startup.cs#L184)):*

> ```
>             // If an admin user has been created via the commandline, treat that as a single action and shut the
>             // server down so the calling script knows it's been completed successfully.
> ```

Again, I think that this should be part of normal startup.  Having the backend create the user and then exit it as a "single action" means that instead of a single, well-defined function, now the backend has two different modes - am I creating a user or am I running the backend?
</blockquote></details>

Done.

PR #440, opened 2020-06-16
Refactor Tree Tiles

### LINKED ISSUE #424: Give More Description to TreeView Side Arrows
Right now, the user only sees numbers which seem to represent a system set up by us that the user doesn't have a clue how to read. Having a corresponding label or section name in addition or instead of the numbers would be helpful
image

imnasnainaec:
Resolves #424

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/440)


StevesBro:
___
*[src/resources/tree/index.js, line 10 at r1](https://reviewable.io/reviews/sillsdev/thecombine/440#-MA2cni07tT2HrKKD4iX:-MA2cni1EKI7-obOZlJN:b-ola5ff) ([raw file](https://github.com/sillsdev/thecombine/blob/fc264187d4ee9b78e18fdbd7e811d1339c29476b/src/resources/tree/index.js#L10)):*
> ```JavaScript
> import teeUpLeft from "../../resources/tree/teeUpLeft.svg";
> import teeUpRight from "../../resources/tree/teeUpRight.svg";
> ```

The "../../resources/tree" isn't needed in any of these file paths.

Issue #427, opened 2020-06-12
Create Docker-based Deployment


johnthagen:
Follow on of #414.

This issue is to collect the work that needs to be done to deploy to a production server. The existing deployment process should be able to be greatly simplified now that Docker has been delegated much of the application installation and configuration process.

### Tasks that need to be done

> This list is a work-in-progress

- [ ] Install Docker on the target if not already installed.
  - https://docs.docker.com/engine/install/ubuntu/
- [ ] Download TheCombine source project
  - Option one
    - Install Git
    - `git clone https://github.com/sillsdev/TheCombine.git`
  - Option two
    - `curl https://github.com/sillsdev/TheCombine/archive/master.zip master.zip`
    - Extract
- [ ] Generate/configure `.env.web`
  - [ ] Generate a secure, random `ASPNETCORE_JWT_SECRET_KEY`
- [ ] Generate [`production.yml` overlay](https://docs.docker.com/compose/production/#modify-your-compose-file-for-production) with `nginx.build.args` (Example added in #446)
  - [ ] Disable `CAPTCHA_REQUIRED` if needed
  - [ ] Generate a new `CAPTCHA_SITE`?
- [ ] Generate/pull in SSL certificates from Lets Encrypt
  - [ ] [Install SSL certificates into Docker Container](https://github.com/sillsdev/TheCombine/blob/01aaeb8c6c920b0bdcd561e33dccd686f14804b9/README.md#ssl-certificates)
- [ ] `docker compose up --build -d`

All persistant data (Mongo database, SSL certs) will be stored in persistent Docker volumes associated with the Docker Compose project.

jmgrady:
A couple of comments/questions:

1. Don't you mean install docker on the target?  Both host & target?  Ansible is a good option for installing on the target.
2. Regarding downloading the source - what about using a tag/branch specifier - tags for released builds, branches for testing.  `curl` is probably better than `git` so that we don't have the entire repo.
3. .env.web should be for the backend, correct?  Not the web (nginx) container.  Maybe I read that wrong.  The current deployment sets ASPNETCORE_JWT_SECRET_KEY as an environment variable for the backend service.
[...]

PR #446, opened 2020-06-19
Use identical baseURL in production

### LINKED ISSUE #431: Add frontend build configuration to use identical URL for API
Spin off task from #415.

It would simplify configuration and deployment if the frontend had a build configuration for the `baseURL` that resulted in something like:

```ts
if (buildConfig.apiIdenticalUrl) {
    return `${same_base_url_as_window}/v1`;
} else {
  // Do what is currently done.
}
```

This would allow us to not have to specify an exact URL ahead of time for each host we build for.

johnthagen:
Closes #431
Addresses the last part of this comment: https://github.com/sillsdev/TheCombine/issues/427#issuecomment-644373042

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/446)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 
 # Production
 /build
+production.yml
 
 # Misc
 .DS_Store
--- a/README.md
+++ b/README.md
@@ -138,8 +138,8 @@ database objects), create a user normally and then execute:
 
 Install [Docker](https://docs.docker.com/get-docker/).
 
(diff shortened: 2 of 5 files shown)


johnthagen:
@jmgrady This is ready for review. Production configuration should be greatly simplified now.

- You no longer need to manually specify the exact URL that the target will be built on
- The README explains how to override a small subset of the build args to disable the captcha. Ansible should be able to generate this file easily on deployment prior to running the build if this needs to be disabled.

PR #476, opened 2020-07-03
Improve code coverage of tests using PermissionServiceMock

johnthagen:
Also refactor `IPermissionService.HasProjectPermission()` argument order to be consistent with other interface methods.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/476)


johnthagen:
___
*[Backend.Tests/UserRoleControllerTests.cs, line 105 at r6](https://reviewable.io/reviews/sillsdev/thecombine/476#-MCTGL_KANXfaO1NHMdc:-MCTIi4w8bMU2OHL31uh:bb9yu2p) ([raw file](https://github.com/sillsdev/thecombine/blob/f37aa5dfcfee2862aa06254b9e212013f7c63d28/Backend.Tests/UserRoleControllerTests.cs#L105)):*

Since `var userRole = RandomUserRole();` and `var userRole = _userRoleService.Create(RandomUserRole()).Result;` are used in so many of the tests, could they be defined once globally (with two different names) and used everywhere, as with `_projId`?
</blockquote></details>

I think the issue is that in some tests such as `TestGetAllUserRoles` we are specifically counting the number of user roles created, so if we create one in `Setup`, I think it would make the tests a little less isolated?

PR #497, opened 2020-07-08
Re-enable recording from data entry

### LINKED ISSUE #472: Unable to add audio clip during data entry
If I create a new project and add an entry, the audio record button doesn't work for me.

audio-import-issue

But when I do it in the Data Cleanup menu, the button works.

audio-import-works

It also seems like my gloss wasn't saved from the entry?

imnasnainaec:
Resolves #472

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/497)


imnasnainaec:
Currently only the last recording is saved.
A) Do we want all recordings blindly saved?
B) Do we want an interactive PronunciationsCell to be able to play/delete recordings same as in ReviewEntries?

PR #505, opened 2020-07-12
Update Environment Variable Names

### LINKED ISSUE #501: Rename backend environment variable prefix
@jmgrady Pointed out in a PR review that the current `ASPNETCORE_` prefixes for the backend environment variables are a bit confusing, especially in the context of a larger system configuration.

I propose we rename the environment variables to an app-specific prefix, `COMBINE_`, and use this convention for all environment variables various parts of the application needs set.

To avoid conflicts, we should wait for #413 and #445 to be merged before doing this as they both interact with backend env variables.

johnthagen:
Closes #501 

In addition:

- No longer check in .env files, but create a template for developers to copy from to avoid the risk of checked in files being used in production

**Note**: Once this is merged into `master`, developers will need to update their environment variables locally.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/505)


johnthagen:
___
*[.env.backend.template, line 8 at r1](https://reviewable.io/reviews/sillsdev/thecombine/505#-MCCgDEh7_WMPWI6l7Kv:-MCRWfsY5PvnDD3LoYJT:bjv4ol2) ([raw file](https://github.com/sillsdev/thecombine/blob/91ccd6837bd987e09722ee64ece5d71dc51ee267/.env.backend.template#L8)):*

Actually, the SMTP configuration items should not be stored in source control - at least in plaintext.  Especially the username and password.
</blockquote></details>

Yeah, we don't want to check in the `SMTP` variables as you've said. The standard way to deal with this for developers is to create a template like this, and the developers copy it one time and fill it in. That file is ignored in git to avoid the possibility it accidentally gets checked in (which is a real risk that I've seen happen accidentally at work).

PR #527, opened 2020-07-20
Require usernames to be at least 3 characters long.

### LINKED ISSUE #525: Project Settings - Find Users won't pick up users with 1 or 2 character usernames
We either need to force the user to enter a username at least 3 characters long or modify Find Users to display exact matches for one and two character long usernames.

imnasnainaec:
Resolves #525

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/527)


imnasnainaec:
Still need to write more unit tests for various registration form combinations.

Issue #451, opened 2020-06-24
Site Settings needs user administration


imnasnainaec:
The Site Settings page (only available to administrative users) has an empty Users section for administering all users on the site. That section needs functionality added.

![image](https://user-images.githubusercontent.com/6411521/85620309-98b6f000-b631-11ea-8e13-84f53050ed4d.png)

imnasnainaec:
(This issue is only available once pr #310 is merged into master.)

PR #539, opened 2020-07-21
Site Settings User List

14JAW:
Allows Site Admin to view and delete users. It provides a Modal to confirm its deletion. In a later pull request, we should probably provide a page for the Admin to edit details of all the users.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/539)


imnasnainaec:
Resolves #451

Issue #437, opened 2020-06-15
Update to MongoDB 4.2


jmgrady:
Current deployment scripts install MondoDB Community Edition Server.  Update the deployment scripts to install 4.2 (includes updating the repo keys and the apt sources).

jmgrady:
Closed by PR #550 Docker Deploy.

Issue #378, opened 2020-06-01
Add instructions for using Windows to install TheCombine on a target machine


jmgrady:
Replace the vagrant-installer vm with instructions for installing TheCombine from the Windows Subsystem for Linux.  (Ansible is used to install and maintain TheCombine on the target hardware but Ansible only runs on Linux.)  This will allow users who need to update any target to run it from a Windows 10 PC.

jmgrady:
Closed by PR #550 Docker Deploy.

PR #550, opened 2020-07-27
Docker deploy

### LINKED ISSUE #344: Create initial database admin on target systems

jmgrady:
Create scripts to deploy TheCombine as a set of Docker Containers

Closes #344, #378, #427, #437 

This PR implements the Ansible playbooks to deploy TheCombine to QA & Production machines with TheCombine running inside docker containers.  `./docs/docker_deploy/README.md` contains the instructions for using the playbooks as well as a description of their design.

Caveat: These playbooks only work for the QA server or the live server; NUCs are not supported yet.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/550)


johnthagen:
___
*[nginx/conf.d/thecombine.conf, line 3 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_06Qn3SlbTMcgxw0w:-MD_06QoEb8-xQafLBt5:bz6qdph) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/nginx/conf.d/thecombine.conf#L3)):*
> ```
> server {
>   listen              443 ssl;
>   server_name	        ${SERVER_NAME};
> ```

Very minor, but could remove two spaces to align this value with the others nearby.

johnthagen:
Is the plan that we can remove the `deploy` folder once we've fully moved to Docker-based deployment (to avoid duplication)?

johnthagen:
___
*[docs/docker_deploy/README.md, line 117 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_5LjQ1Ol_iQy1PxNX:-MD_5LjR3TaRYarCWepz:b-vhewsx) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L117)):*
> ```Markdown
>     3. install Docker Engine
>     4. install Docker Compose (including a link from `/usr/local/bin` to the executable)
>   2. Create a combine user - the combine user will be used for installing, building and running *TheCombine*.  The default user name is `combine`.
> ```

Should this be 3?

johnthagen:
___
*[docs/docker_deploy/README.md, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_5jYu2dMsdQzLGkix:-MD_5jYu2dMsdQzLGkiy:bvsyjmr) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L1)):*
> ```Markdown
> # Deploy *TheCombine* In Docker Containers
> ```

Minor, but if the lines in this file were wrapped (to say, 120 characters) it would be easier to open the file next to another file on the screen. We wrap the source code lines at a certain length for this reason too.

johnthagen:
___
*[docker_deploy/roles/docker_install/tasks/main.yml, line 51 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_KN_O-UGog98igdtJ:-MD_KN_O-UGog98igdtK:b-thprr8) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/docker_install/tasks/main.yml#L51)):*
> ```YAML
> - name: Install Docker Compose
>   get_url:
>     url: "https://github.com/docker/compose/releases/download/1.26.0/docker-compose-{{ ansible_system }}-{{ ansible_architecture }}"
> ```

I noticed that Compose 1.26.2 is now available.

Perhaps in our release process we should add a checklist of packages to update versions on so we don't forget to bump these versions in the future?

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 14 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_SfpPDhIy9GJkwnd1:-MD_SfpPDhIy9GJkwnd2:bqmjkvk) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L14)):*
> ```
>         - COMBINE_CAPTCHA_REQUIRED={{ config_captcha_required }}
>         - COMBINE_CAPTCHA_SITE={{ config_captcha_sitekey }}
>     volumes:
> ```

These `volumes` are listed in the base `docker-compose.yml` file, so we can simplify this overlay and remove them from here.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 33 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_T4FRBTSIydRr62qE:-MD_T4FRBTSIydRr62qF:b-5objxl) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L33)):*
> ```
>     restart: unless-stopped
> 
> volumes:
> ```

The `letsencrypt` `volume` is already in the base `docker-compose.yml` so we should be able to remove it here to simplify this file.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 27 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_TFCt1h0W-BfhXKWP:-MD_TFCt1h0W-BfhXKWQ:b-ug3osl) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L27)):*
> ```
>       - "/bin/bash"
>       - "{{ frontend_start_script }}"
>     depends_on:
> ```

`depends_on` section is already in the base `docker-compose.yml` so it can be removed here.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 10 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_TgYo0OTF8F4G4Yw0:-MD_TgYo0OTF8F4G4Yw1:b-n2vgow) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L10)):*
> ```
>   frontend:
>     build:
>       context: .
> ```

`context` is already in the base `docker-compose.yml` and can be removed from here to simplify this file.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 23 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_V2T30UHBWRAPwuep:-MD_V2T30UHBWRAPwueq:b-cmmbl5) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L23)):*
> ```
>         SSL_CERTIFICATE: {{ ssl_certificate }}
>         SSL_PRIVATE_KEY: {{ ssl_private_key }}
>         BACKEND_URL: "http://backend"
> ```

I think `BACKEND_URL` should be removed from this list? The default port we want to run the backend on now is `5000`, so we can eventually run the backend as a non-root user. The base `docker-compose.yml` has this set correctly, so I don't think there is any need to override this (to port `80`).

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 4 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_VjDJ9svefKAJSNjG:-MD_VjDJ9svefKAJSNjH:brvhiry) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L4)):*
> ```
> services:
>   backend:
>     entrypoint:
> ```

Is the default `ENTRYPOINT` not working properly for deployment? For simplicity, it would be nice if we could use a common entrypoint and avoid duplication of the `dotnet BackFramework.dll` command.Perhaps we can modify the `launch.sh` if something isn't working as needed.

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 13 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_jSqP8bpcDG_n9gAJ:-MD_jSqP8bpcDG_n9gAK:blk0ejt) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/defaults/main.yml#L13)):*
> ```YAML
>   ASPNETCORE_ENVIRONMENT: Production
>   DOTNET_PRINT_TELEMETRY_MESSAGE: false
>   ASPNETCORE_URLS: http://+
> ```

Is there a reason to use port `80` instead of `5000`? This is all within the Docker Compose project, so I think we can use port `5000` consistently in all configurations and not need to do any overrides.

johnthagen:
___
*[docker_deploy/roles/combine_init/templates/docker-compose.init.yml.j2, line 6 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_kGhQDyaBDocGfhbQ:-MD_kGhQDyaBDocGfhbR:b-pfpqp6) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_init/templates/docker-compose.init.yml.j2#L6)):*
> ```
>     env_file:
>       - .env.backend.init
>     entrypoint:
> ```

I wonder if we can use the common `ENTRYPOINT` for this and place the admin name into the `.env` file. Then we could simplify this overlay.

johnthagen:
___
*[docs/docker_deploy/README.md, line 61 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_m1CL9iJLZ_kn5vft:-MD_m1CL9iJLZ_kn5vfu:bjlebzn) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L61)):*
> ```Markdown
> 
> To install and start up *TheCombine* you will need to run the following Ansible playbooks.  Each time you will be prompted for passwords:
>  * `BECOME password` - enter your `sudo` password on for the *\<target_user\>* on the *\<target\>* machine.
> ```

...password on for the... -> ...password for the...

johnthagen:
___
*[docs/docker_deploy/README.md, line 113 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDa52o75u4pSoYsSADa:-MDa52o75u4pSoYsSADb:b-yiy5y2) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L113)):*
> ```Markdown
>   1. Install Ansible dependencies to allow subsequent playbooks
>   2. Install the docker subsystem.  This includes:
>     1. install docker prerequisite packages
> ```

This section is not rendering properly in GitHub. Perhaps due to `:` or only using two spaces rather than 4?

johnthagen:
___
*[docs/docker_deploy/README.md, line 132 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDa5LPrCZ0MmnO9zDWy:-MDa5LPrCZ0MmnO9zDWz:b-prxjlv) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L132)):*
> ```Markdown
>   1. install *TheCombine* source files from the *GitHub* repository.  It will install the `master` branch unless the `combine_source_version` variable specifies a different branch.  The source files are installed in `/home/combine/src/combine`.
>   2. configure the docker containers for the *\<target\>* system:
>     1.  creates the `docker-compose.prod.yml` file to override specific defaults in `docker-compose.yml`.
> ```

Several of these sections also don't render properly in GitHub.

johnthagen:
Not sure why this is happening, but the `generate_config.sh` script is not being applied properly. I published the code to a VM using ansible, but I the frontend isn't properly built with `useConnectionBaseUrlForApi: true,` for some reason. This causes connection errors.

You can see in the network monitor that it is trying to go to `https://localhost:5001` instead of `https://target`. Perhaps there is something different about how the `staticfloat/nginx-certbot` image works.

image

johnthagen:
Strange, when I run the docker-compose project locally (outside Ansible/VMs) the frontend container is exiting. I thought this was working when I first started looking at this branch, which is odd.

The line it is hanging at is: https://github.com/staticfloat/docker-nginx-certbot/blob/25041145ef1b5c505978c6966652f5c2bc9b1214/src/scripts/entrypoint.sh#L48

```
docker-compose logs -f
Attaching to thecombine_frontend_1, thecombine_backend_1, thecombine_database_1
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_SMTP_SERVER` is not defined. Email services will not work.
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_SMTP_PORT` is not defined. Email services will not work.
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_SMTP_USERNAME` is not defined. Email services will not work.
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_SMTP_PASSWORD` is not defined. Email services will not work.
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_SMTP_ADDRESS` is not defined. Email services will not work.
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_SMTP_FROM` is not defined. Email services will not work.
backend_1   | fail: BackendFramework.Startup[0]
backend_1   |       Environment variable: `COMBINE_PASSWORD_RESET_EXPIRE_TIME` is not defined. Using default value: 15
backend_1   | info: BackendFramework.Startup[0]
backend_1   |       No admin user name provided, skipped admin creation
backend_1   | warn: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
backend_1   |       Storing keys in a directory '/root/.aspnet/DataProtection-Keys' that may not be persisted outside of the container. Protected data will be unavailable when container is destroyed.
backend_1   | info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[0]
backend_1   |       User profile is available. Using '/root/.aspnet/DataProtection-Keys' as key repository; keys will not be encrypted at rest.
backend_1   | info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[58]
backend_1   |       Creating key {d978786f-cae2-4565-9cf3-5c3842f88b56} with creation date 2020-07-31 19:42:20Z, activation date 2020-07-31 19:42:20Z, and expiration date 2020-10-29 19:42:20Z.
backend_1   | warn: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[35]
backend_1   |       No XML encryptor configured. Key {d978786f-cae2-4565-9cf3-5c3842f88b56} may be persisted to storage in unencrypted form.
backend_1   | info: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[39]
backend_1   |       Writing data to file '/root/.aspnet/DataProtection-Keys/key-d978786f-cae2-4565-9cf3-5c3842f88b56.xml'.
backend_1   | Hosting environment: Production
backend_1   | Content root path: /app
backend_1   | Now listening on: http://[::]:5000
backend_1   | Application started. Press Ctrl+C to shut down.
database_1  | 2020-07-31T19:42:18.395+0000 I  CONTROL  [main] Automatically disabling TLS 1.0, to force-enable TLS 1.0 specify --sslDisabledProtocols 'none'
database_1  | 2020-07-31T19:42:18.397+0000 W  ASIO     [main] No TransportLayer configured during NetworkInterface startup
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] MongoDB starting : pid=1 port=27017 dbpath=/data/db 64-bit host=07b419ebecad
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] db version v4.2.8
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] git version: 43d25964249164d76d5e04dd6cf38f6111e21f5f
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] OpenSSL version: OpenSSL 1.1.1  11 Sep 2018
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] allocator: tcmalloc
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] modules: none
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] build environment:
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten]     distmod: ubuntu1804
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten]     distarch: x86_64
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten]     target_arch: x86_64
database_1  | 2020-07-31T19:42:18.398+0000 I  CONTROL  [initandlisten] options: { net: { bindIp: "*" } }
database_1  | 2020-07-31T19:42:18.399+0000 I  STORAGE  [initandlisten]
database_1  | 2020-07-31T19:42:18.399+0000 I  STORAGE  [initandlisten] ** WARNING: Using the XFS filesystem is strongly recommended with the WiredTiger storage engine
database_1  | 2020-07-31T19:42:18.399+0000 I  STORAGE  [initandlisten] **          See http://dochub.mongodb.org/core/prodnotes-filesystem
database_1  | 2020-07-31T19:42:18.399+0000 I  STORAGE  [initandlisten] wiredtiger_open config: create,cache_size=12297M,cache_overflow=(file_max=0M),session_max=33000,eviction=(threads_min=4,threads_max=4),config_base=false,statistics=(fast),log=(enabled=true,archive=true,path=journal,compressor=snappy),file_manager=(close_idle_time=100000,close_scan_interval=10,close_handle_minimum=250),statistics_log=(wait=0),verbose=[recovery_progress,checkpoint_progress],
database_1  | 2020-07-31T19:42:19.100+0000 I  STORAGE  [initandlisten] WiredTiger message [1596224539:100143][1:0x7f12b6c69b00], txn-recover: Set global recovery timestamp: (0, 0)
database_1  | 2020-07-31T19:42:19.179+0000 I  RECOVERY [initandlisten] WiredTiger recoveryTimestamp. Ts: Timestamp(0, 0)
database_1  | 2020-07-31T19:42:19.239+0000 I  STORAGE  [initandlisten] Timestamp monitor starting
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten]
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten] ** WARNING: Access control is not enabled for the database.
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten] **          Read and write access to data and configuration is unrestricted.
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten]
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten]
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten] ** WARNING: /sys/kernel/mm/transparent_hugepage/enabled is 'always'.
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten] **        We suggest setting it to 'never'
database_1  | 2020-07-31T19:42:19.290+0000 I  CONTROL  [initandlisten]
database_1  | 2020-07-31T19:42:19.291+0000 I  STORAGE  [initandlisten] createCollection: admin.system.version with provided UUID: 5ff09ea8-5b95-40c2-a821-8d9eb61bcfe0 and options: { uuid: UUID("5ff09ea8-5b95-40c2-a821-8d9eb61bcfe0") }
database_1  | 2020-07-31T19:42:19.372+0000 I  INDEX    [initandlisten] index build: done building index _id_ on ns admin.system.version
database_1  | 2020-07-31T19:42:19.372+0000 I  SHARDING [initandlisten] Marking collection admin.system.version as collection version: <unsharded>
database_1  | 2020-07-31T19:42:19.372+0000 I  COMMAND  [initandlisten] setting featureCompatibilityVersion to 4.2
database_1  | 2020-07-31T19:42:19.373+0000 I  SHARDING [initandlisten] Marking collection local.system.replset as collection version: <unsharded>
database_1  | 2020-07-31T19:42:19.373+0000 I  STORAGE  [initandlisten] Flow Control is enabled on this deployment.
database_1  | 2020-07-31T19:42:19.373+0000 I  SHARDING [initandlisten] Marking collection admin.system.roles as collection version: <unsharded>
database_1  | 2020-07-31T19:42:19.373+0000 I  STORAGE  [initandlisten] createCollection: local.startup_log with generated UUID: f1464bc6-1581-437a-b026-ef2e536d66f9 and options: { capped: true, size: 10485760 }
database_1  | 2020-07-31T19:42:19.424+0000 I  INDEX    [initandlisten] index build: done building index _id_ on ns local.startup_log
database_1  | 2020-07-31T19:42:19.424+0000 I  SHARDING [initandlisten] Marking collection local.startup_log as collection version: <unsharded>
database_1  | 2020-07-31T19:42:19.425+0000 I  FTDC     [initandlisten] Initializing full-time diagnostic data capture with directory '/data/db/diagnostic.data'
database_1  | 2020-07-31T19:42:19.426+0000 I  SHARDING [LogicalSessionCacheRefresh] Marking collection config.system.sessions as collection version: <unsharded>
database_1  | 2020-07-31T19:42:19.427+0000 I  CONTROL  [LogicalSessionCacheReap] Sessions collection is not set up; waiting until next sessions reap interval: config.system.sessions does not exist
database_1  | 2020-07-31T19:42:19.427+0000 I  STORAGE  [LogicalSessionCacheRefresh] createCollection: config.system.sessions with provided UUID: 501e07f1-b20f-4acd-abb0-adf9713e014c and options: { uuid: UUID("501e07f1-b20f-4acd-abb0-adf9713e014c") }
database_1  | 2020-07-31T19:42:19.427+0000 I  NETWORK  [listener] Listening on /tmp/mongodb-27017.sock
database_1  | 2020-07-31T19:42:19.427+0000 I  NETWORK  [listener] Listening on 0.0.0.0
database_1  | 2020-07-31T19:42:19.427+0000 I  NETWORK  [listener] waiting for connections on port 27017
database_1  | 2020-07-31T19:42:19.478+0000 I  INDEX    [LogicalSessionCacheRefresh] index build: done building index _id_ on ns config.system.sessions
database_1  | 2020-07-31T19:42:19.534+0000 I  INDEX    [LogicalSessionCacheRefresh] index build: starting on config.system.sessions properties: { v: 2, key: { lastUse: 1 }, name: "lsidTTLIndex", ns: "config.system.sessions", expireAfterSeconds: 1800 } using method: Hybrid
database_1  | 2020-07-31T19:42:19.534+0000 I  INDEX    [LogicalSessionCacheRefresh] build may temporarily use up to 200 megabytes of RAM
database_1  | 2020-07-31T19:42:19.534+0000 I  INDEX    [LogicalSessionCacheRefresh] index build: collection scan done. scanned 0 total records in 0 seconds
database_1  | 2020-07-31T19:42:19.534+0000 I  INDEX    [LogicalSessionCacheRefresh] index build: inserted 0 keys from external sorter into index in 0 seconds
database_1  | 2020-07-31T19:42:19.543+0000 I  INDEX    [LogicalSessionCacheRefresh] index build: done building index lsidTTLIndex on ns config.system.sessions
database_1  | 2020-07-31T19:42:19.552+0000 I  COMMAND  [LogicalSessionCacheRefresh] command config.system.sessions command: createIndexes { createIndexes: "system.sessions", indexes: [ { key: { lastUse: 1 }, name: "lsidTTLIndex", expireAfterSeconds: 1800 } ], $db: "config" } numYields:0 reslen:114 locks:{ ParallelBatchWriterMode: { acquireCount: { r: 2 } }, ReplicationStateTransition: { acquireCount: { w: 3 } }, Global: { acquireCount: { r: 1, w: 2 } }, Database: { acquireCount: { r: 1, w: 2, W: 1 } }, Collection: { acquireCount: { r: 4, w: 1, R: 1, W: 2 } }, Mutex: { acquireCount: { r: 3 } } } flowControl:{ acquireCount: 1 } storage:{} protocol:op_msg 125ms
database_1  | 2020-07-31T19:42:20.000+0000 I  SHARDING [ftdc] Marking collection local.oplog.rs as collection version: <unsharded>
database_1  | 2020-07-31T19:42:20.019+0000 I  NETWORK  [listener] connection accepted from 172.24.0.3:46766 #1 (1 connection now open)
database_1  | 2020-07-31T19:42:20.050+0000 I  NETWORK  [conn1] received client metadata from 172.24.0.3:46766 conn1: { driver: { name: "mongo-csharp-driver", version: "2.10.4.0" }, os: { type: "Linux", name: "Linux 4.19.104-microsoft-standard #1 SMP Wed Feb 19 06:37:35 UTC 2020", architecture: "x86_64", version: "4.19.104-microsoft-standard" }, platform: ".NET Core 3.1.5" }
frontend_1  | templating scripts from /etc/nginx/user.conf.d to /etc/nginx/conf.d
frontend_1  | Substituting variables $SERVER_NAME $SSL_CERTIFICATE $SSL_PRIVATE_KEY $BACKEND_URL
frontend_1  |  -> /etc/nginx/user.conf.d/thecombine.conf
frontend_1  | Done with startup
frontend_1  | Run certbot
frontend_1  | CERTBOT_EMAIL environment variable undefined; certbot will do nothing
frontend_1  | /scripts/entrypoint.sh: line 48:    17 Hangup                  nginx -g "daemon off;"
thecombine_frontend_1 exited with code 0
```

johnthagen:
On the Ansible deployment, I went into the container and looked at `config.js` and it looks generated properly:

```
# ls -l
total 36
-rw-r--r-- 1 root root  494 May 26 15:00 50x.html
-rw-r--r-- 1 root root 1938 Jul 31 18:44 asset-manifest.json
-rw-r--r-- 1 root root  158 Jul 31 18:44 config.js
-rw-r--r-- 1 root root 3667 Jul 31 18:43 favicon.ico
-rw-r--r-- 1 root root 2137 Jul 31 18:44 index.html
-rw-r--r-- 1 root root  301 Jul 31 18:43 manifest.json
-rw-r--r-- 1 root root 1988 Jul 31 18:44 precache-manifest.7e1c8f38a488adafb47a3802921a7336.js
-rw-r--r-- 1 root root 1181 Jul 31 18:44 service-worker.js
drwxr-xr-x 5 root root 4096 Jul 31 18:44 static
# cat config.js	
window['runtimeConfig'] = {
    useConnectionBaseUrlForApi: true,
    captchaRequired: True,
    captchaSiteKey: "6Le6BL0UAAAAAMjSs1nINeB5hqDZ4m3mMg3k67x3"
}
```

So, I'm at a loss at the moment.

johnthagen:
___
*[nginx/nginx.conf, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDaVxgNAjPhYbi99jCp:-MDaVxgNAjPhYbi99jCq:b-wp1gn2) ([raw file](https://github.com/sillsdev/thecombine/blob/07a16cbb98516a4e82d9cb79b29f44daa660f5fa/nginx/nginx.conf#L1)):*
> ```Nginx
> # Redirect all HTTP requests to HTTPS.
> ```

We need to retain this section so that users that browse to `http://<URL>` it will redirect to `https://<URL>`

johnthagen:
___
*[nginx/nginx.conf, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDaVxgNAjPhYbi99jCp:-MDaZ40e93uv1xVKdrtN:bx801g7) ([raw file](https://github.com/sillsdev/thecombine/blob/07a16cbb98516a4e82d9cb79b29f44daa660f5fa/nginx/nginx.conf#L1)):*

We need to retain this section so that users that browse to `http://<URL>` it will redirect to `https://<URL>`
</blockquote></details>

Based on our conversation, we could just add a comment that docker-nginx-certbot is taking care of this for us.

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 13 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_jSqP8bpcDG_n9gAJ:-MDdu3yr6CMQhA-Vp2Fp:b-2jyjws) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/defaults/main.yml#L13)):*

Actually, I thought that is something you set up.  I was wondering the same thing.  I probably misunderstood something.  I will change to use port 5000.
</blockquote></details>

The default port for ASP.NET is 80 if you pass nothing, but port 5000 is convention for HTTP. This is already set in the `docker-compose.yml` now so Ansible should not need to change it at all. We can define that we're using 5000 once and leave it constant. Again, this port is only internal to the Docker compose project so it does not get exposed to the outside world (like clients).

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 4 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_VjDJ9svefKAJSNjG:-MDdu_8q9n8-_ODSFAf2:b3lknln) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L4)):*

Done.

It seemed cleaner and more straight-forward to me to call the method directly.  It seems like an odd side-effect to call that method if the username is defined.  It also obviates the need for the environment variable and the launch.sh.

I am fine with setting the environment variable as well.
</blockquote></details>

I agree it is somewhat odd looking, but since the way to affect runtime behavior with Docker is through env variables, that was the only universal way I could find to control the behavior from the outside with the outside not having to know about the internals of the container (names of DLLs, etc.).

One nice side effect is that we can unify all configuration of how the containers get launched through one set of controls. For example, as a developer running Docker without Ansible, I can create a user in the same way as ansible (using the env variables).

johnthagen:
___
*[docker_deploy/roles/docker_install/tasks/main.yml, line 51 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_KN_O-UGog98igdtJ:-MDdvT5S7RVn7IhAh85Q:b-yfb2xe) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/docker_install/tasks/main.yml#L51)):*

I can add that as a variable at a high level.  That way we can check/update in one place - perhaps ./docker_deploy/vars/packages.yml.
I would prefer it if it was included in an Ubuntu repo though!
</blockquote></details>

Yeah, making it a var would make it more noticable. I do not understand why Docker Compose is included on Windows and macOS and not Linux.
You have that really nice `docs/docker_deploy/README.md`, perhaps you could add a section at the bottom that talks about "dependency version bumping".

johnthagen:
___
*[docs/docker_deploy/README.md, line 117 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_5LjQ1Ol_iQy1PxNX:-MDdvs861vqzAc_6MfST:bxexggy) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L117)):*

Normally markdown automatically numbers correctly but I will fix this because it makes the source easier to read.  I also need fix the indentation because many of the netsed lists in this section do not display properly.
</blockquote></details>

Yeah, Markdown renderers will render it correctly. Another alternative I've seen a few places is to just use `1.` for the whole list and let the renderers count up. I think numbering it ourselves does make the raw files easier to read though.

johnthagen:
___
*[docs/docker_deploy/README.md, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_5jYu2dMsdQzLGkix:-MDdw8QX6dk4yOFij5rj:b87qe6d) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docs/docker_deploy/README.md#L1)):*

Done.
However, what editor are you using to view the file?  Atom wraps the lines very intelligently for smaller windows or viewing files side-by-side.
</blockquote></details>

I'm using JetBrains Rider but this also happens in VSCode if you don't have word wrap on. I find that when working with source code, word wrap makes it harder to understand the control flow so I don't have that enabled.

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 13 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_jSqP8bpcDG_n9gAJ:-MDdwzxg4qr3KYvkFNY0:bu7ym5t) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/defaults/main.yml#L13)):*

The default port for ASP.NET is 80 if you pass nothing, but port 5000 is convention for HTTP. This is already set in the `docker-compose.yml` now so Ansible should not need to change it at all. We can define that we're using 5000 once and leave it constant. Again, this port is only internal to the Docker compose project so it does not get exposed to the outside world (like clients).
</blockquote></details>

To be clear, I propose that we:

- In Backend/Dockerfile, leave `ENV ASPNETCORE_URLS=http://+:5000`
- In docker-compose.yml, leave `BACKEND_URL: "http://backend:5000"`
- Remove all references to any overrides that Ansible makes for `BACKEND_URL` or `ASPNETCORE_URLS` as this value will never need to be modified.

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 13 at r1](https://reviewable.io/reviews/sillsdev/thecombine/550#-MD_jSqP8bpcDG_n9gAJ:-MDeIx1x8gXoB6OvwzuY:b-8wxhpn) ([raw file](https://github.com/sillsdev/thecombine/blob/973f05cf68cd4f774e55ff8fb7af672373c69785/docker_deploy/roles/combine_containers/defaults/main.yml#L13)):*

I was planning on
 * removing the definition of `BACKEND_URL` in  docker-compose.yml
 * remove `BACKEND_URL`from `ENVSUBST_VARS` list of environment variables to substitute in the nginx configuration file
 * restoring `proxy_pass    http://backend:5000;` in nginx/conf.d/thecombine.conf
 
</blockquote></details>

Yes, removing the configuration of `BACKEND_URL` and hardcoding into the NGINX conf also works. That is good since we won't need to configure it. I wasn't sure if there had been another reason you wanted to expose the `BACKEND_URL` configuration.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 19 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeKlgE3sON-Ds97QAV:-MDeKlgE3sON-Ds97QAW:b-km0upg) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L19)):*
> ```
>         SSL_CERTIFICATE: {{ ssl_certificate }}
>         SSL_PRIVATE_KEY: {{ ssl_private_key }}
>     command:
> ```

Is this necessary? The default command for the staticfloat/nginx-certbot image is already set to this. See https://github.com/staticfloat/docker-nginx-certbot/blob/25041145ef1b5c505978c6966652f5c2bc9b1214/src/Dockerfile#L32

Perhaps we can just remove this and can just use the default `CMD`?

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 4 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeL5SpCnoAy4SBtMGy:-MDeL5SpCnoAy4SBtMGz:b-igut73) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L4)):*
> ```
> services:
>   backend:
>     entrypoint:
> ```

Per previous comments, can this `entrypoint` be removed so that all Docker invocations use the same, default `ENTRYPOINT`?
I do think that in the future we might be able to simplify/improve the default `ENTRYPOINT`, but as our first step it would be nice to not duplicate this launching information.

johnthagen:
___
*[docker_deploy/roles/combine_shortcuts/tasks/main.yml, line 36 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeLmgi9s_lDDbsfiYV:-MDeLmgi9s_lDDbsfiYW:b3nsvjf) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docker_deploy/roles/combine_shortcuts/tasks/main.yml#L36)):*
> ```YAML
>       alias combine-build="docker-compose -f docker-compose.yml -f {{ production_override_yaml }} build --parallel"
>       alias combine-down="docker-compose -f docker-compose.yml -f {{ production_override_yaml }} down"
>       alias combine-logs="docker-compose -f docker-compose.yml -f {{ production_override_yaml }} logs"
> ```

Just as a note, usually I run `docker-compose logs --follow` to get real time streaming updates. I'm not 100% sure how these shortcuts are envisioned to be used, but just wanted to double check it was intentional to leave off `--follow`.

johnthagen:
___
*[docker_deploy/roles/combine_shortcuts/tasks/main.yml, line 35 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeLyI1Dk-wEpw7uJGC:-MDeLyI1Dk-wEpw7uJGD:b-9eqpsg) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docker_deploy/roles/combine_shortcuts/tasks/main.yml#L35)):*
> ```YAML
>       alias combine-up="docker-compose -f docker-compose.yml -f {{ production_override_yaml }} up --detach"
>       alias combine-build="docker-compose -f docker-compose.yml -f {{ production_override_yaml }} build --parallel"
>       alias combine-down="docker-compose -f docker-compose.yml -f {{ production_override_yaml }} down"
> ```

Another small note, **if** we want to provide a command to totally wipe out the database (for example if there is an API breaking change that we can't trivially fix with a mongo command), we could add another alias that does the equivalent of `docker-compose down --volumes`. Maybe we would call it `combine-erase`.

johnthagen:
___
*[docker_deploy/vars/packages.yml, line 9 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeMQYT8mtBvKjNz3t4:-MDeMQYT8mtBvKjNz3t5:b-96zzz0) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docker_deploy/vars/packages.yml#L9)):*
> ```YAML
> 
> # Packages installed by playbook_docker.yml
> docker_compose_version: 1.26.0
> ```

Can we bump this to 1.26.2? https://github.com/docker/compose/releases

johnthagen:
___
*[docs/docker_deploy/README.md, line 98 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeOgemDtVKtFpSkeE3:-MDeOgen1RDiH-38BWwj:b-eagix5) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docs/docker_deploy/README.md#L98)):*
> ```Markdown
>       own inventory file (see [below](#creating-your-own-inventory-file)).
> 
>  * `ansible-playbook playbook_publish.yml --limit <target> -u combine --ask-vault-pass`
> ```

Could we add a new heading above this line? I think something that initially threw me off was that this command is other "the first playbook" and "Installing Combine Pre-requsites" but is really the step that actually installs the application.

Maybe add some new text above this line:

```
#### Install Combine

Run the following playbook to install and start TheCombine
```

johnthagen:
___
*[docs/docker_deploy/README.md, line 1 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeQ-2K98q0mgdFrjd8:-MDeQ-2K98q0mgdFrjd9:bdez2kw) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docs/docker_deploy/README.md#L1)):*
> ```Markdown
> # Deploy *TheCombine* In Docker Containers
> ```

Should we have a short "System Requirements" section? I remember you said this didn't work on 512MB of memory. Perhaps we should list some minimum and recommended specs as best we understand them?

johnthagen:
___
*[docker_deploy/roles/combine_init/tasks/main.yml, line 22 at r2](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDeT7Hr6_-fu_kks17O:-MDeT7Hr6_-fu_kks17P:b-f9o8vc) ([raw file](https://github.com/sillsdev/thecombine/blob/f9ab94a699a64c2d84f384e103957381aecc8cd8/docker_deploy/roles/combine_init/tasks/main.yml#L22)):*
> ```YAML
> 
> - name: run the initialization
>   command: docker-compose -f docker-compose.yml -f docker-compose.init.yml up backend
> ```

For me, when I tried out the latest branch the publish step hung on initialization.

```
$ ansible-playbook playbook_publish.yml --limit qa -u combine --ask-vault-pass -i dev.hosts.yml 
Vault password: 

PLAY [Setup TheCombine Using Docker] ****************************************************************

TASK [Gathering Facts] ******************************************************************************
[DEPRECATION WARNING]: Distribution Ubuntu 18.04 on host target should use /usr/bin/python3, but is 
using /usr/bin/python for backward compatibility with prior Ansible releases. A future Ansible 
release will default to using the discovered platform python for this host. See 
https://docs.ansible.com/ansible/2.9/reference_appendices/interpreter_discovery.html for more 
information. This feature will be removed in version 2.12. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.
ok: [target]

TASK [combine_source : get source code from GitHub] *************************************************
changed: [target]

TASK [combine_containers : create production docker-compose file] ***********************************
changed: [target]

TASK [combine_containers : create backend .env file] ************************************************
changed: [target]

TASK [combine_containers : build the containers] ****************************************************
changed: [target]

TASK [combine_init : create initialization docker-compose file] *************************************
changed: [target]

TASK [combine_init : create backend init .env file] *************************************************
changed: [target]

TASK [combine_init : run the initialization] ********************************************************
```

On the target machine, it looks like the admin user name didn't get set.

```
user@target:/home/combine/src/combine$ sudo docker-compose logs -f
[sudo] password for user: 
Attaching to combine_backend_1, combine_database_1
backend_1   | info: BackendFramework.Startup[0]
backend_1   |       No admin user name provided, skipped admin creation
backend_1   | warn: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
backend_1   |       Storing keys in a directory '/root/.aspnet/DataProtection-Keys' that may not be persisted outside of the container. Protected data will be unavailable when container is destroyed.
backend_1   | info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[0]
backend_1   |       User profile is available. Using '/root/.aspnet/DataProtection-Keys' as key repository; keys will not be encrypted at rest.
backend_1   | info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[58]
backend_1   |       Creating key {418a9eb6-16a6-45f7-af6f-56e237a4d4ad} with creation date 2020-08-01 14:37:19Z, activation date 2020-08-01 14:37:19Z, and expiration date 2020-10-30 14:37:19Z.
backend_1   | warn: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[35]
backend_1   |       No XML encryptor configured. Key {418a9eb6-16a6-45f7-af6f-56e237a4d4ad} may be persisted to storage in unencrypted form.
backend_1   | info: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[39]
backend_1   |       Writing data to file '/root/.aspnet/DataProtection-Keys/key-418a9eb6-16a6-45f7-af6f-56e237a4d4ad.xml'.
backend_1   | Hosting environment: Production
backend_1   | Content root path: /app
backend_1   | Now listening on: http://[::]:5000
backend_1   | Application started. Press Ctrl+C to shut down.
```

Does `COMBINE_ADMIN_USERNAME` need to get set for this step so the entrypoint picks it up?

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 11 at r3](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDfbV-v6Tc2ipgt-a0n:-MDfbV-v6Tc2ipgt-a0o:b-b80jus) ([raw file](https://github.com/sillsdev/thecombine/blob/4c4f53c0986daefbcadd75116d025a69fcd75183/docker_deploy/roles/combine_containers/defaults/main.yml#L11)):*
> ```YAML
> # for non-development deployments
> combine_env_vars:
>   ASPNETCORE_ENVIRONMENT: Production
> ```

Let's move this into `backend/Dockerfile`, that way all use cases using that image are built for production (as is the other Docker images).

In that file:

```
ENV ASPNETCORE_ENVIRONMENT=Production
```

This also should be removed from `the_combine_app/defaults/main.yml` then as well.

This help keep the default `docker-compose` environment consistent across all use cases.

johnthagen:
I am still having issues browsing to the QA server I pushed to a VM. The backend URL that is being used is `https://localhost:5001` even though I see a correct config.js come back. I have cleared the browser cache and also tried disabling cacheing.





My `dev.hosts.yml`:

```
---

all:
  hosts:
  children:
    nuc:
      hosts:
        nuc2:
    server:
      hosts:
        thecombine:
    qa:
      hosts:
        target:
          combine_server_name: target
          combine_captcha_required: "false"
          combine_captcha_sitekey:
          combine_source_version: docker_deploy
```

jmgrady:
I'll try it out tomorrow afternoon.  One thing I have found helpful for connecting to VMs is to add a network adapter and set it up for bridged mode  - this is work Virtual Box - you're using VMWare, though, correct?

⁣Jim

"In practice, the difference between theory and practice is greater than it is in theory."​

jmgrady:
Will do. Good idea.

⁣Jim

"In practice, the difference between theory and practice is greater than it is in theory."​

johnthagen:
> One thing I have found helpful for connecting to VMs is to add a network adapter and set it up for bridged mode  - this is work Virtual Box - you're using VMWare, though, correct?

Yes, VMWare Workstation on Windows. The VMs can communicate fine as best I can tell (ping, HTTPS, etc all work). What I'm seeing is bizarre as I can't think of a reason why it's happening.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 16 at r3](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDgJifQ0dSSQAQSfInA:-MDgJifQ0dSSQAQSfInB:bxeizxt) ([raw file](https://github.com/sillsdev/thecombine/blob/4c4f53c0986daefbcadd75116d025a69fcd75183/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L16)):*
> ```
>         SSL_CERTIFICATE: {{ ssl_certificate }}
>         SSL_PRIVATE_KEY: {{ ssl_private_key }}
>     restart: unless-stopped
> ```

I discovered what the issue was about the `docker-compose` not working natively from the source checkout without Ansible. The frontend keeps restarting until NGINX is successfully found. That is why this works in production, but not locally (this requires `restart: unless_stopped`).

As a short term fix, we should move this `restart: unless-stopped` into the based `docker-compose.yml`, along with probably the `database` one as well. This is a good example of why it's good to have a little difference between the base Docker config and production as possible because this config was actually necessary for the Docker Compose project to work at all using the new certbot image.

As as longer term effort, we should figure out why the frontend keeps closing and needs to be restarted and solve the underlying issue (perhaps the certbot image is not configured correctly, or there is something else we need to do). It seems unlikely we should have to use `restart: unless-stopped` just to make it start up properly.

johnthagen:
___
*[docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2, line 16 at r3](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDgJifQ0dSSQAQSfInA:-MDgKfyfDeVavOIhV5-U:b-jkmu9x) ([raw file](https://github.com/sillsdev/thecombine/blob/4c4f53c0986daefbcadd75116d025a69fcd75183/docker_deploy/roles/combine_containers/templates/docker-compose.prod.yml.j2#L16)):*
> of why it's good to have a little difference between the base Docker config

of why it's good to have **as** little difference between the the base Docker config

johnthagen:
I do not see the the incorrect backend URL issue when I run the docker compose project vanilla straight from the source checkout without the Ansible publish (after manually fixing the `restart: unless-stopped` issue I reported earlier). I'm guessing there must be something about how I'm using the `dev.hosts.yml` or similar config that is messing with the frontend build/hosting.

johnthagen:
See my `dev.hosts.yml` in this comment: https://github.com/sillsdev/TheCombine/pull/550#issuecomment-667581346

johnthagen:
Also, for what it's worth I was getting an error shown on the login page where the captcha normally is with this `dev.hosts.yml`. Perhaps related?

johnthagen:
> How do you currently manage this at your work?

Anytime we do a new publish, we should afterwards run `docker image prune --force` to clean up any old images that are dangling. This will remove the replaced images that are no longer accessible. It is safe in that any named images will not be removed, so the newest versions of our containers we just installed will not be removed.

johnthagen:
Here is the error I get in the frontend using this `dev.hosts.yml`:

```
---

all:
  hosts:
  children:
    nuc:
      hosts:
        nuc2:
    server:
      hosts:
        thecombine:
    qa:
      hosts:
        target:
          combine_server_name: target
          combine_captcha_required: "false"
          combine_captcha_sitekey:
          combine_source_version: docker_deploy
```

johnthagen:
___
*[docker-compose.yml, line 13 at r5](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDjCoWg4XQuxQgYuB_Y:-MDjCoWg4XQuxQgYuB_Z:bqbajyp) ([raw file](https://github.com/sillsdev/thecombine/blob/23acfe85e8257f1269b90b3546fccab2a0ded084/docker-compose.yml#L13)):*
> ```YAML
>     env_file:
>       - ./.env.backend
>     restart: unless-stopped
> ```

I need to test this, but just need to confirm that if you create an admin user, this doesn't keep restarting the backend infinitely.

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 4 at r5](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDjE3ma-KaRuRTYAFDy:-MDjE3ma-KaRuRTYAFDz:biqbkpa) ([raw file](https://github.com/sillsdev/thecombine/blob/23acfe85e8257f1269b90b3546fccab2a0ded084/docker_deploy/roles/combine_containers/defaults/main.yml#L4)):*
> ```YAML
> #######################################
> # Default configs for development
> config_captcha_required: true
> ```

Here, the value is `true`, where in other files it is `"true"`. I think this quoting might be part of the `config.js` underlying issues.

When it gets to Javascript, we don't want to quotes, so we need to verify what's going on.

Other files:

- `qa-thecombine.psonet.yml`
- `thecombine.yml`

johnthagen:
Having issues with the publish Jinja2 template after the latest update.

```
$ ansible-playbook playbook_publish.yml --limit qa -u combine --ask-vault-pass -i dev.hosts.yml
Vault password: 

PLAY [Setup TheCombine Using Docker] ****************************************************************

TASK [Gathering Facts] ******************************************************************************
[DEPRECATION WARNING]: Distribution Ubuntu 18.04 on host target should use /usr/bin/python3, but is 
using /usr/bin/python for backward compatibility with prior Ansible releases. A future Ansible 
release will default to using the discovered platform python for this host. See 
https://docs.ansible.com/ansible/2.9/reference_appendices/interpreter_discovery.html for more 
information. This feature will be removed in version 2.12. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.
ok: [target]

TASK [combine_source : get source code from GitHub] *************************************************
changed: [target]

TASK [combine_containers : create production docker-compose file] ***********************************
changed: [target]

TASK [combine_containers : create backend .env file] ************************************************
fatal: [target]: FAILED! => {"changed": false, "msg": "AnsibleError: Unexpected templating type error occurred on (﻿{% for item in combine_env_vars|dict2items %}\n{{ item.key }}={{ item.value }}\n{% endfor %}\n{% for item in combine_private_env_vars|dict2items %}\n{{ item.key }}={{ item.value }}\n{% endfor %}\n): dict2items requires a dictionary, got <type 'NoneType'> instead."}

PLAY RECAP ******************************************************************************************
target                     : ok=3    changed=2    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

```

`dev.hosts.yml`:

```
---

all:
  hosts:
  children:
    nuc:
      hosts:
        nuc2:
    server:
      hosts:
        thecombine:
    qa:
      hosts:
        target:
          combine_server_name: target
          combine_captcha_required: "false"
          combine_source_version: docker_deploy
```

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 4 at r5](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDjE3ma-KaRuRTYAFDy:-MDjIww0CaqC5QR8lb0f:bt08973) ([raw file](https://github.com/sillsdev/thecombine/blob/23acfe85e8257f1269b90b3546fccab2a0ded084/docker_deploy/roles/combine_containers/defaults/main.yml#L4)):*

Here, the value is `true`, where in other files it is `"true"`. I think this quoting might be part of the `config.js` underlying issues.

When it gets to Javascript, we don't want to quotes, so we need to verify what's going on.

Other files:

- `qa-thecombine.psonet.yml`
- `thecombine.yml`
</blockquote></details>

Confirmed a bug related to this. When I use  `"false"` in my `dev.hosts.yml` configuration that is being translated to `True` in config.js.

See: https://github.com/sillsdev/TheCombine/pull/550#issuecomment-667581346

`captchaRequired: True,` rather than `true`.

johnthagen:
___
*[docs/docker_deploy/README.md, line 132 at r5](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDjJQMg-F6DaEqK8dVN:-MDjJQMg-F6DaEqK8dVO:bmh5lll) ([raw file](https://github.com/sillsdev/thecombine/blob/23acfe85e8257f1269b90b3546fccab2a0ded084/docs/docker_deploy/README.md#L132)):*
> ```Markdown
>      - config_captcha_sitekey:
> 
>     config_captcha_required should be "true" or "false" (including the quotes);
> ```

I believe this line in the docs may be incorrect and the source of the bugs I was seeing. I think `"true"` and `"false"` are not being translated properly to Javascript `boolean`s. I believe the quotes must not be there (or else somewhere along the line it's being interpreted as a `string`.

johnthagen:
___
*[docs/docker_deploy/README.md, line 132 at r5](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDjJQMg-F6DaEqK8dVN:-MDjJsjOA6lIjF1_qy91:b-n1u0r5) ([raw file](https://github.com/sillsdev/thecombine/blob/23acfe85e8257f1269b90b3546fccab2a0ded084/docs/docker_deploy/README.md#L132)):*

I believe this line in the docs may be incorrect and the source of the bugs I was seeing. I think `"true"` and `"false"` are not being translated properly to Javascript `boolean`s. I believe the quotes must not be there (or else somewhere along the line it's being interpreted as a `string`.
</blockquote></details>

See https://github.com/sillsdev/TheCombine/pull/550#issuecomment-667672962

johnthagen:
___
*[docker_deploy/roles/combine_containers/defaults/main.yml, line 4 at r5](https://reviewable.io/reviews/sillsdev/thecombine/550#-MDjE3ma-KaRuRTYAFDy:-MDjxYoYC0yFSxE66SaC:bg1m2f4) ([raw file](https://github.com/sillsdev/thecombine/blob/23acfe85e8257f1269b90b3546fccab2a0ded084/docker_deploy/roles/combine_containers/defaults/main.yml#L4)):*

I have another branch where I am testing doing all of this for the NUC (using the self-signed certs for now).  The group_vars/nuc/main.yml file contains:
```
---
#################################################
# Group specific configuration items
#
# Group: nuc
#################################################

config_captcha_required: "false"
config_captcha_sitekey: ""

...
```
The `config.js` that is created is:
```
root@db4581f6d7a5:/usr/share/nginx/html# cat config.js
window['runtimeConfig'] = {
    useConnectionBaseUrlForApi: true,
    captchaRequired: false,
    captchaSiteKey: ""
}
```
So I think that is coming from someplace else.  I'll let you know if I find anything more.
</blockquote></details>

For what it's worth, Ansible in built on Python and in Python the keyword for true is `True`, which seems like what leaked through in my setup (Javascript it is `true`).

This is further complicated in that YAML itself does all kinds of implicit boolean conversions itself, so you could very well be right that the quotes are a red herring.

docker_deploy/roles/combine_config/templates/docker-compose.yml.j2
```diff
@@ -1,37 +1,41 @@
...
-      context: .
-      args:
-        - COMBINE_CAPTCHA_REQUIRED=true
-        - COMBINE_CAPTCHA_SITE=6Le6BL0UAAAAAMjSs1nINeB5hqDZ4m3mMg3k67x3
+    image: {{ combine_image_frontend }}
+    env_file:
+      - ./.env.frontend
     volumes:
+      - ./nginx/conf.d:/etc/nginx/user.conf.d:ro
+      - ./nginx/scripts:/usr/share/nginx/html/scripts:ro
```
johnthagen, line 20: Would it be simpler to `COPY` these files in the `Dockerfile` than rely on Docker Compose for this? That would make the `Dockerfile` more stand-alone and keep all of the `COPY` logic in one place.

jmgrady: Possibly for the nginx/conf.d but not for the scripts.  The scripts folder is for the config.js which is a runtime configuration.  If we put that in the build step, then the container is not reusable between different target types (QA server/Live server/NUC).

johnthagen: Resolved in 8443cc1dabd3e155dcc149514a4f1b8abd8d8ba4

johnthagen:
___
*[docker_deploy/roles/combine_config/templates/docker-compose.yml.j2, line 20 at r11](https://reviewable.io/reviews/sillsdev/thecombine/550#-MGtZ5CfEE2nztFq5bcA-r11-20:-MGyLIl2-CJTueZWR81S:bb6un1j) ([raw file](https://github.com/sillsdev/thecombine/blob/ca1317be8ef904d87c521af060e628aa2a0ff317/docker_deploy/roles/combine_config/templates/docker-compose.yml.j2#L20)):*

Resolved in 8443cc1dabd3e155dcc149514a4f1b8abd8d8ba4
</blockquote></details>

Done

johnthagen (commented):
> Reviewed 3 of 3 files at r4, 2 of 2 files at r5, 1 of 39 files at r6, 2 of 8 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/550)* status: 17 of 39 files reviewed, 1 unresolved discussion (waiting on @jmgrady and @johnthagen)

johnthagen:
___
*[docker_deploy/roles/package_install/defaults/main.yml, line 5 at r13](https://reviewable.io/reviews/sillsdev/thecombine/550#-MGyceIeB63nYjEppmjO:-MGyceIeB63nYjEppmjP:bk44hkg) ([raw file](https://github.com/sillsdev/thecombine/blob/4144149cf2067caa670600838ac91ef358648cd1/docker_deploy/roles/package_install/defaults/main.yml#L5)):*
> ```YAML
> required_packages:
>   - python3
>   - python3-pip
> ```

What is `pip` used for in the deployment environment? Are there extra PyPI packages that are needed outside of what comes with Ansible?

johnthagen:
___
*[docker_deploy/vars/config_common.yml, line 11 at r13](https://reviewable.io/reviews/sillsdev/thecombine/550#-MGycvZU13oi92APK4ep:-MGycvZU13oi92APK4eq:b-qvx8z) ([raw file](https://github.com/sillsdev/thecombine/blob/4144149cf2067caa670600838ac91ef358648cd1/docker_deploy/vars/config_common.yml#L11)):*
> ```YAML
> frontend_env_dir: "{{ source_dir }}"
> 
> production_override_yaml: docker-compose.prod.yml
> ```

Now that we are templating out the `docker-compose.yml` file. Can we remove this `.prod` override?

johnthagen (commented):
> Reviewed 14 of 39 files at r6, 1 of 1 files at r8, 4 of 8 files at r12, 3 of 3 files at r13, 3 of 3 files at r14.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/550)* status: all files reviewed, 2 unresolved discussions (waiting on @jmgrady)

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/550)* status: 36 of 39 files reviewed, 1 unresolved discussion (waiting on @johnthagen)

___
*[docker_deploy/vars/config_common.yml, line 11 at r13](https://reviewable.io/reviews/sillsdev/thecombine/550#-MGycvZU13oi92APK4ep:-MGyltYy7nQiqWUnxwhV:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/4144149cf2067caa670600838ac91ef358648cd1/docker_deploy/vars/config_common.yml#L11)):*

Now that we are templating out the `docker-compose.yml` file. Can we remove this `.prod` override?
</blockquote></details>

Done.

johnthagen (approved):
> Reviewed 1 of 1 files at r15, 2 of 2 files at r16.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/550)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #554, opened 2020-07-30
UniqueProjectNames

StevesBro:
Requires all project names on TheCombine to be unique

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/554)


StevesBro:
___
*[Backend/Controllers/ProjectController.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/thecombine/554#-MDX0u519eiXOc8Ff6ui:-MD_pzSj7tOjiE4981_Q:b5wmrwv) ([raw file](https://github.com/sillsdev/thecombine/blob/07a16cbb98516a4e82d9cb79b29f44daa660f5fa/Backend/Controllers/ProjectController.cs#L43)):*

> ```
>             if (!_permissionService.HasProjectPermission(HttpContext, Permission.DatabaseAdmin))
>             {
>                 return new ForbidResult();
>             }
> ```
We may want to relax the permissions instead of removing this check entirely.
</blockquote></details>

createProject  (Post) in the ProjectController doesn't check for permissions so any user can create a project regardless of permissions.

StevesBro:
___
*[Backend/Controllers/ProjectController.cs, line 43 at r1](https://reviewable.io/reviews/sillsdev/thecombine/554#-MDX0u519eiXOc8Ff6ui:-MD_ttlv1dI8tFscMHyZ:b-q0mmfk) ([raw file](https://github.com/sillsdev/thecombine/blob/07a16cbb98516a4e82d9cb79b29f44daa660f5fa/Backend/Controllers/ProjectController.cs#L43)):*
<details><summary><i>Previously, 14JAW wrote…</i></summary><blockquote>

Right. In the Post case we want any user to be able to create one but we only want the admin to be able to access all projects. Rather than removing permissions, would it be possible to move the duplicate check to the backend? That way the user knows that there is a project with that name but has no way of accessing or modifying it.
</blockquote></details>

Excellent idea. I can create a isDuplicateProject function that will just return a bool.

StevesBro:
___
*[src/components/ProjectScreen/CreateProject/CreateProjectComponent.tsx, line 37 at r3](https://reviewable.io/reviews/sillsdev/thecombine/554#-MDaDwBt49LTpQNYC7xH:-MDaELWW-wM5P7LoCWJa:bji8de1) ([raw file](https://github.com/sillsdev/thecombine/blob/9b3dca19c1ff5d26b557aac436adc7aea2a5f5e7/src/components/ProjectScreen/CreateProject/CreateProjectComponent.tsx#L37)):*

Need to reconcile `error: { name: boolean };` and `error: { empty: boolean; nameTaken: boolean };` in the state.
</blockquote></details>

Ya, I'm having to resolve merge conflicts

Issue #549, opened 2020-07-25
Allow user to see Entire Font Characters


14JAW:
For some Characters, the current display will cut off part of the bottom or something to where it does not look like the character we desire.

imnasnainaec:
Test word: g̱wa̱g̱wax̱wa̱lał

imnasnainaec:
`TextField` cuts off diacritics below tails. Issue submitted: https://github.com/mui-org/material-ui/issues/22104

imnasnainaec:
Was only partially fixed by #560.

imnasnainaec:
A taller text field has been merged into material-ui for future release: https://github.com/mui-org/material-ui/pull/22149

imnasnainaec:
It is in an alpha release: https://github.com/mui-org/material-ui/releases/tag/v5.0.0-alpha.7

PR #560, opened 2020-07-31
Include more fonts for languages

14JAW:
1. To add a new font: 
- In `types/theme.tsx` --> "fontFamily: ['"<Enter name of desired font here>"','"Noto Sans"','"Roboto"','"Helvetica"','"Arial"',"sans-serif",].join(","), 
 --- Each font in the list is used based on what operating system the user is on. Starting from the left-most font, the program looks through to see if the character desired is included in the font. If not, then it continues searching through the next font.
-In 'public/index.html' --> Include a link to the desired font. If using Google Fonts, select the font and font style and then click the embed tab to the right. It will have a <Link > component for you to copy and paste in the <header > <right here > </header>.
2. For the Cards, I adjusted the characters to the center to avoid any cut-off characters. A state "fontHeight" is set to a constant that should be tall enough to have any characters. State "cardWidth" is able to be set in CharacterListComponent. To make a card have a set-width, edit the style "maxWidth" to "width".

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/560)


imnasnainaec:
This is only a partial resolution to #549  .

PR #570, opened 2020-08-03
Consolidate local storage

imnasnainaec:
Consolidating and cleaning up use of localStorage before undertaking using it for audio with DataEntry > NewEntry.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/570)


imnasnainaec:
User properties like `password` and `isAdmin` shouldn't be in localStorage, and when user is fetched from localStorage, often only the user.id is needed. This refactor only stores the user properties needed by various parts of the combine (`id`, `name`, `token`, `workedProjects`).

PR #586, opened 2020-08-12
Make Username and Email Case-Insensitive

StevesBro:
I make username and email ToLower whenever it is sent to a controller.

### DIFF
--- a/Backend/Controllers/UserController.cs
+++ b/Backend/Controllers/UserController.cs
@@ -37,7 +37,7 @@ namespace BackendFramework.Controllers
         [HttpPost("forgot")]
         public async Task<IActionResult> ResetPasswordRequest([FromBody] PasswordResetData data)
         {
-            var email = data.Email;
+            var email = data.Email.ToLower();
             // create password reset
             var resetRequest = await _passwordResetService.CreatePasswordReset(email);
 
@@ -116,6 +116,7 @@ namespace BackendFramework.Controllers
         {
             try
             {
+                cred.Username = cred.Username.ToLower();
                 var user = await _userService.Authenticate(cred.Username, cred.Password);
                 if (user == null)
                 {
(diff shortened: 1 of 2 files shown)


StevesBro:
___
*[Backend/Controllers/UserController.cs, line 40 at r2](https://reviewable.io/reviews/sillsdev/thecombine/586#-MEdSx8uDk8W2TN6RCqB:-MEdiyIA2Qys9sMicpCX:bp6i8xv) ([raw file](https://github.com/sillsdev/thecombine/blob/6eff655bd7c70d174d5cca34280eb36e2dcdc49a/Backend/Controllers/UserController.cs#L40)):*

I recommend using ToLowerInvariant()
There are some weird edge cases where the result could depend on the localization language.
We aren't localizing the backend yet, but we will need to in the future in order to get translations of the invitation e-mails.
</blockquote></details>

Gotcha, I've switched em.

johnthagen:
___
*[Backend/Controllers/UserController.cs, line 40 at r3](https://reviewable.io/reviews/sillsdev/thecombine/586#-MEgt2Ld0PaYHGMJotWe:-MEgt2Ld0PaYHGMJotWf:b-bud27q) ([raw file](https://github.com/sillsdev/thecombine/blob/7177133a80c420f83331ea3d4c7101a027bc27cb/Backend/Controllers/UserController.cs#L40)):*
> ```Smalltalk
>         public async Task<IActionResult> ResetPasswordRequest([FromBody] PasswordResetData data)
>         {
>             var email = data.Email.ToLowerInvariant();
> ```

In the backend unit tests,  `Util.RandString()` is used to create user names. The implementation of that only uses the characters `a` through `z` (lower case).

This means we aren't currently testing that lower casing is being applied in the `UserController`. Could the unit tests be updated to test that lower-casing is being applied?

StevesBro:
___
*[Backend/Controllers/UserController.cs, line 40 at r3](https://reviewable.io/reviews/sillsdev/thecombine/586#-MEgt2Ld0PaYHGMJotWe:-MF7ROc5BPq4mKrbgi05:b-fv6skg) ([raw file](https://github.com/sillsdev/thecombine/blob/ea6f6354ea7689e84ee6a27ac565a9bb2b5ab357/Backend/Controllers/UserController.cs#L40)):*

We'll address this unit test in a follow up PR.
</blockquote></details>

Currently working on tests right now

PR #619, opened 2020-08-19
Hide deleted senses in data entry

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/619)


imnasnainaec:
Something similar needs to be done for MergeDup, but that is proving more difficult.

PR #631, opened 2020-08-21
[ProjectSettings] Make project languages editable

imnasnainaec:
Component in place, but update project errors, possibly a race condition:
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/631)


imnasnainaec:
I was thinking the languages should be changeable in case a user (accidentally) selects an analysis language that doesn't match the lift import, in which case the senses wouldn't be visible in DataEntry > ExistingDataTable nor in ReviewEntries. However, it may be better to just allow adding more Analysis languages and switching which one is first in the list (i.e., active).

imnasnainaec:
We don't wan't to edit the languages. We only want: https://github.com/sillsdev/TheCombine/issues/712

PR #637, opened 2020-08-24
ExportLIFT gets url instead of file from backend

imnasnainaec:
This doesn't doesn't play nice with local absolute paths. and I'm not sure how to test its behavior on a remotely hosted TheCombine. In my test, the .zip file is located at `/home/dror/.CombineFiles/5f4001464d14445cedb7dc9e/Export/LiftExportCompressed-5f4001464d14445cedb7dc9e_2020-08-24_03-38-42.zip` but the frontend strips of the initial `\` and treats it as a relative path.
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/637)


imnasnainaec:
Fixed instead by #638

PR #643, opened 2020-08-25
Do not export deleted senses

jasonleenaylor:
* Exclude non-active senses during the export
* Minor code clean-up

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/643)

### DIFF
--- a/Backend.Tests/LiftControllerTests.cs
+++ b/Backend.Tests/LiftControllerTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using BackendFramework.Controllers;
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
___
*[Backend/Services/LiftApiServices.cs, line 149 at r1](https://reviewable.io/reviews/sillsdev/thecombine/643#-MFanWYU4uPA-fd3WzJb:-MFapGtwDcaimXmP3Nzt:b-1kzr27) ([raw file](https://github.com/sillsdev/thecombine/blob/5a8466e2c8ba07456d7f10c94a464cbe2ed93905/Backend/Services/LiftApiServices.cs#L149)):*

Shouldn't this be `!activeWords.Contains(x)` to get the words that aren't in the frontier?
</blockquote></details>

Um, yes indeed.

PR #658, opened 2020-08-27
Move NewEntry above recent entries and scroll up on update

### LINKED ISSUE #609: [DataEntryComponent] Bad scrolling when ExistingDataTable is too long
When ExistingDataTable is too long, past the bottom, hitting enter in NewEntry causes unhelpful down-scrolling.

This is likely due to the app bar covering the content that we are trying to scroll into view.

- [ ] Make ExistingDataTable its own scrollable window when it’s taller than its container

imnasnainaec:
Resolves #609

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/658)


imnasnainaec:
Having the NewEntry below was and still is intended.

PR #659, opened 2020-08-27
Allow escape from VernDialog

imnasnainaec:
Currently you are forced to make a selection if the VernDialog pops up.
By enabling escape, somebody who wants to change the vernacular instead of making a selection can do so.

Also new: Escape in the second dialog (sense selection) retracts the user's entry selection and returns them to the first dialog for another go at entry selection.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/659)


StevesBro:
___
*[src/components/DataEntry/DataEntryTable/VernWithSuggestions/VernDialog/VernDialog.tsx, line 35 at r2](https://reviewable.io/reviews/sillsdev/thecombine/659#-MFpsNuFAHA5ntqTlRfl:-MFpsNuFAHA5ntqTlRfm:b-6y87l0) ([raw file](https://github.com/sillsdev/thecombine/blob/3dda98e55301bc257776cf99a2c1fd2d8dd7c3e1/src/components/DataEntry/DataEntryTable/VernWithSuggestions/VernDialog/VernDialog.tsx#L35)):*
> ```XML
>       onClose={() => props.handleClose()}
>       disableBackdropClick
>       disableEscapeKeyDown
> ```

It would be nice if we could use the escape key for "Select a Sense" also.

imnasnainaec:
> _[src/components/DataEntry/DataEntryTable/VernWithSuggestions/VernDialog/VernDialog.tsx, line 35 at r2](https://reviewable.io/reviews/sillsdev/thecombine/659#-MFpsNuFAHA5ntqTlRfl:-MFpsNuFAHA5ntqTlRfm:b-6y87l0) ([raw file](https://github.com/sillsdev/thecombine/blob/3dda98e55301bc257776cf99a2c1fd2d8dd7c3e1/src/components/DataEntry/DataEntryTable/VernWithSuggestions/VernDialog/VernDialog.tsx#L35)):_
> 
> > ```
> >       onClose={() => props.handleClose()}
> >       disableBackdropClick
> >       disableEscapeKeyDown
> > ```
> 
> It would be nice if we could use the escape key for "Select a Sense" also.

That might be nice, yes, but we need to decide what escaping out of the second dialog (sense selection) does:
1. Assume "add new sense"
2. Clear out entry selection (from first dialog) and close the dialogs
3. Clear out entry selection and return to first dialog to make a new selection (or press escape again)

StevesBro:
___
*[src/components/DataEntry/DataEntryTable/VernWithSuggestions/VernDialog/VernDialog.tsx, line 35 at r2](https://reviewable.io/reviews/sillsdev/thecombine/659#-MFpsNuFAHA5ntqTlRfl:-MFq2QHh3hl4lLCLQU_J:brumfxl) ([raw file](https://github.com/sillsdev/thecombine/blob/3dda98e55301bc257776cf99a2c1fd2d8dd7c3e1/src/components/DataEntry/DataEntryTable/VernWithSuggestions/VernDialog/VernDialog.tsx#L35)):*

That might be nice, yes, but we need to decide what escaping out of the second dialog (sense selection) does:
1. Assume "add new sense"
2. Clear out entry selection (from first dialog) and close the dialogs
3. Clear out entry selection and return to first dialog to make a new selection (or press escape again) 
</blockquote></details>

I think those are all good options. I would prefer options 2.

PR #649, opened 2020-08-26
Update RecordRTC

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/649)

PR #661, opened 2020-08-28
Sync up @types/recordrtc package-lock.json with npm install

johnthagen:
I believe this update was accidentally missed in #649

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/661)


johnthagen:
@imnasnainaec Could you double check this is intended? I think there was just a missing `npm install` run before you pushed up your final commits in #649, but want to confirm with you.

PR #666, opened 2020-08-28
Add backend audio merging in MergeDuplicates

StevesBro:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/666)


imnasnainaec:
@jasonleenaylor One situation to consider:
* A has 1 sense and no recordings, B has 2 senses and 1 recording
* Move one of the senses from B to A in MergeDups
* Now A has 2 senses and 1 recording, B has 1 sense and 1 recording
* A and B have copies of the same recording with the same filename (but in different locations so you can delete one and the other stays)
* Export the project and only one copy of that filename is exported and the lift file points both entries at the same audio file
* This imports in into FieldWorks without a problem, but now there are two words pointing to the same audio file in FieldWorks

imnasnainaec:
> One situation to consider:
> * A has 1 sense and no recordings, B has 2 senses and 1 recording
> * Move one of the senses from B to A in MergeDups
> * Now A has 2 senses and 1 recording, B has 1 sense and 1 recording
> [...]

Recordings are no longer duplicated. Now the oddest known case is:
* A has 2 senses and 1 recording
* Move one of A's senses to B and another to C at the same time, causing A to be removed
* One of B or C will inherit A's recording, whichever merges last, which isn't clear to the user

Issue #349, opened 2020-05-29
There is no easy way to jump back to top of semantic tree


stevendyk:
[5/26 requirements #9](https://docs.google.com/document/d/1A9Wv7SY_peIqdfGEBgzdl-7Pn19Dn70CDy9apVOUiak/edit#heading=h.ktdfqtvquehx): There isn’t an easy way to get out of a node, back up to the top of the tree, if I want to just start over with my navigation

stevendyk:
will be complete when #7 is done

imnasnainaec:
Perhaps clicking on Data Entry should do that when already in Data Entry, but we'd still need a method when selecting a domain in Review Entries.

imnasnainaec:
In both cases, maybe ESCAPE should return somebody to the top of the tree, and ESCAPE from the top of the tree in Review Entries should exit from TreeView.

jasonleenaylor:
Clicking on Data Entry is the development plan for solving this issue.

Issue #702, opened 2020-09-08
Domain search is too restrictive


imnasnainaec:
It's not case sensitive, but it is accent sensitive (in Spanish, "limite" doesn't work for "Límite"--the accent is required over the i). For multi-word domain, the whole phrase is needed, so neither "universe" nor "creation" suffices to find "universe, creation"--neither "language" nor "thought" brings up "Language And Thought".

The search is primarily done by the numeric semantic domain id, so this isn't high priority.

Alternatively/Additionally, we may want to show suggestions as the user types.

jasonleenaylor:
If we find that there is a strong use case for domain searching we can create a new issue with the user felt need as the starting requirements.

imnasnainaec:
A user trying to search for domain 2.6.5.1: Man in Arabic used "رجل" rather than "رجُل" and was surprised that nothing came up. Now we have reason to strip off accents/modifiers in the search.

PR #711, opened 2020-09-09
[TreeViewHeader] Navigate back to tree root with 0; allow 2-digit domain numbers; etc.

imnasnainaec:
This refactor improves searching for domains by their id in 3 ways:
* "0" returns to the top of the tree
  - This is a (partial) solution to #349
* Numbers can be more than 1 digit, in case the set of semantic domains expands in the future
  - E.g., currently "1.10" is not a domain, but navigation with the old code would fail even if it were
* Returns the parent if child doesn't exist
  - E.g., "2.2.0" and "2.2.99" both navigate to "2.2"
  - E.g., "18" and "0.3.4.5.7" both navigate to the top of the tree

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/711)

### DIFF
--- a/src/components/TreeView/TreeViewHeader.tsx
+++ b/src/components/TreeView/TreeViewHeader.tsx
@@ -28,11 +28,13 @@ export default class TreeViewHeader extends React.Component<
   TreeHeaderState
 > {
   animating: boolean;
+  numSep: string;
 
   constructor(props: TreeHeaderProps) {
     super(props);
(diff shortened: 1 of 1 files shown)


imnasnainaec:
Related: #349, #702

imnasnainaec:
Need to update the unit tests.

imnasnainaec:
We're going to use ESC to get back up the tree.

PR #742, opened 2020-09-30
Prevent avatar 404 console errors

### LINKED ISSUE #697: Console shows an error when the frontend requests an avatar for a user that hasn't uploaded one
This is particularly an issue in ProjectSettings and SiteSettings where the list of all users is fetched.

imnasnainaec:
- Add property for whether there is an avatar, so that avatars aren't fetched when there isn't one.

Resolves #697

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)


jasonleenaylor (commented):
One more comment requested. And it feels like there needs to be a unit test in here somewhere...

> Reviewed 7 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742#-:-MIzmsJvEgcX2AJmTnHR:bhmqhyt)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[Backend/Services/UserApiServices.cs, line 160 at r2](https://reviewable.io/reviews/sillsdev/thecombine/742#-MIzmNQ_D8jVisuZ_vQ_:-MIzmBIl3oCa7rL-kmcR:b-n0ookm) ([raw file](https://github.com/sillsdev/thecombine/blob/8020fddd6a47e1b2dcd50bb59be8482f8719289b/Backend/Services/UserApiServices.cs#L160)):*
> ```Smalltalk
>             };
>             var token = tokenHandler.CreateToken(tokenDescriptor);
>             Sanitize(user);
> ```

Add a comment here explaining why this method should both Sanitize before the updated AND actually return a user with a token.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)* status: 7 of 8 files reviewed, all discussions resolved (waiting on @imnasnainaec)

___
*[Backend/Services/UserApiServices.cs, line 160 at r2](https://reviewable.io/reviews/sillsdev/thecombine/742#-MIzmNQ_D8jVisuZ_vQ_:-MJ2nLUv2Oh5Kdm3zVVt:bs3mhdg) ([raw file](https://github.com/sillsdev/thecombine/blob/d8618cf0f0dade48e40fdd27b2d8c7d1fbcad485/Backend/Services/UserApiServices.cs#L160)):*

Add a comment here explaining why this method should both Sanitize before the updated AND actually return a user with a token.
</blockquote></details>

Comment added.

johnthagen:
___
*[Backend/Models/User.cs, line 19 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCP8r3Db_BjinkcPNT:-MJCP8r3Db_BjinkcPNU:bnpn97e) ([raw file](https://github.com/sillsdev/thecombine/blob/c63476e9fc671227ea569a19047dd361e5a7fd2d/Backend/Models/User.cs#L19)):*
> ```Smalltalk
> 
>         [BsonElement("hasAvatar")]
>         public Boolean HasAvatar { get; set; }
> ```

Rather than 404'ing if there is no Avatar, in typical REST API's I've used and developed, the backend could simply return JSON `null` for the Avatar to denote "this attribute is empty".

In [OpenAPI](https://apisyouwonthate.com/blog/openapi-v31-and-json-schema-2019-09), you can set the type of an attribute to be `[string, null]`, which means, this can be of type `string` or `null`.

In this example, is it possible to do something like:

If the user has an avatar:

```json
{
  "name": "Alice",
  "avatar": "http://localhost/.../avatar.jpg",
  ...
}
```

And if they do not:

```json
{
  "name": "Bob",
  "avatar": null,
  ...
}
```

In TypeScript we can then just test what is returned from the backend

```typescript
if (user.avatar !== null) {
  ...
}
```

This would could remove the need for a separate implicitly-related field that the client has to check.

johnthagen:
___
*[Backend/Models/User.cs, line 19 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCP8r3Db_BjinkcPNT:-MJCdCp86LHjUaEKRQ8w:bj9wosw) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/Backend/Models/User.cs#L19)):*

Rather than 404'ing if there is no Avatar, in typical REST API's I've used and developed, the backend could simply return JSON `null` for the Avatar to denote "this attribute is empty".

In [OpenAPI](https://apisyouwonthate.com/blog/openapi-v31-and-json-schema-2019-09), you can set the type of an attribute to be `[string, null]`, which means, this can be of type `string` or `null`.

In this example, is it possible to do something like:

If the user has an avatar:

```json
{
  "name": "Alice",
  "avatar": "http://localhost/.../avatar.jpg",
  ...
}
```

And if they do not:

```json
{
  "name": "Bob",
  "avatar": null,
  ...
}
```

In TypeScript we can then just test what is returned from the backend

```typescript
if (user.avatar !== null) {
  ...
}
```

This would could remove the need for a separate implicitly-related field that the client has to check.
</blockquote></details>

When I create a user, currently the avatar is set to `""` instead of `null` in the database:

```json
{"_id":{"$oid":"5f806b9eb6ac3a2c20b90014"},"avatar":"","hasAvatar":false,"name":"u","email":"a@a.com","phone":"","otherConnectionField":"","workedProjects":{},"projectRoles":{},"agreement":false,"password":"xG7RM5IP/DBMWz7fwyT7s/180IP1G13xtVcYijANdxzt5/wC3G47kb+gjnmyYrPm","username":"user","uiLang":"","token":"","isAdmin":false}
```

To express nullness in the type system we could do something like:

```c#
        [BsonElement("avatar")]
        public string? Avatar { get; set; }

        ...
                public User()
        {
            Id = "";
            Avatar = null;
            ...
```

But we'd then want to turn on nullable compiler support project-wide: https://stackoverflow.com/a/61436270

johnthagen:
___
*[Backend/Models/User.cs, line 19 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCP8r3Db_BjinkcPNT:-MJCh1eG4H7k4lMj4WY-:bi8ptcn) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/Backend/Models/User.cs#L19)):*

When I create a user, currently the avatar is set to `""` instead of `null` in the database:

```json
{"_id":{"$oid":"5f806b9eb6ac3a2c20b90014"},"avatar":"","hasAvatar":false,"name":"u","email":"a@a.com","phone":"","otherConnectionField":"","workedProjects":{},"projectRoles":{},"agreement":false,"password":"xG7RM5IP/DBMWz7fwyT7s/180IP1G13xtVcYijANdxzt5/wC3G47kb+gjnmyYrPm","username":"user","uiLang":"","token":"","isAdmin":false}
```

To express nullness in the type system we could do something like:

```c#
        [BsonElement("avatar")]
        public string? Avatar { get; set; }

        ...
                public User()
        {
            Id = "";
            Avatar = null;
            ...
```

But we'd then want to turn on nullable compiler support project-wide: https://stackoverflow.com/a/61436270

</blockquote></details>

Opened a separate issue about tracking nullables project-wide, as that could be addressed separately from this: https://github.com/sillsdev/TheCombine/issues/754

johnthagen:
___
*[Backend/Models/User.cs, line 19 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJChhN1CTCCWQsdSVQj:-MJChhN1CTCCWQsdSVQk:b-56d7dl) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/Backend/Models/User.cs#L19)):*
> ```Smalltalk
> 
>         [BsonElement("hasAvatar")]
>         public Boolean HasAvatar { get; set; }
> ```

This should be declared as

```csharp
public bool HasAvatar { get; set; }
```

for consistency

johnthagen:
___
*[Backend/Services/UserApiServices.cs, line 178 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCkuJo5tPdwK7VoyIE:-MJCkuJo5tPdwK7VoyIF:b8wn406) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/Backend/Services/UserApiServices.cs#L178)):*
> ```Smalltalk
>         {
>             var users = await _userDatabase.Users.Find(_ => true).ToListAsync();
>             users.ForEach(c => Sanitize(c));
> ```

This can be simplified to:

```c#
users.ForEach(Sanitize);
```

johnthagen:
___
*[src/components/ProjectSettings/ProjectUsers/ActiveUsers.tsx, line 39 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCmVRK92n6FrXWSgUQ:-MJCmVRK92n6FrXWSgUR:bqsz6zi) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/src/components/ProjectSettings/ProjectUsers/ActiveUsers.tsx#L39)):*
> ```XML
>   private populateUsers() {
>     getAllUsersInCurrentProject()
>       .then(async (projUsers) => {
> ```

This can be simplified with a `for..of` loop:


```typescript
        this.setState({ projUsers });
        const userAvatar = this.state.userAvatar;
        // for loop rather than .forEach forces each await to finish
        for (let user of projUsers) {
          if (user.hasAvatar) {
            userAvatar[user.id] = await avatarSrc(user.id);
          }
        }
        this.setState({ userAvatar });
```

johnthagen:
___
*[src/components/ProjectSettings/ProjectUsers/ProjectUsers.tsx, line 82 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCneYG5NN_w-pifwpN:-MJCneYG5NN_w-pifwpO:bfcr2yo) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/src/components/ProjectSettings/ProjectUsers/ProjectUsers.tsx#L82)):*
> ```XML
>               ),
>             }));
>             const userAvatar = this.state.userAvatar;
> ```

Can be simplified with `for...of`:

```ts
            const userAvatar = this.state.userAvatar;
            // for loop rather than .forEach forces each await to finish
            for (let user of returnedUsers) {
              if (user.hasAvatar) {
                userAvatar[user.id] = await backend.avatarSrc(user.id);
              }
            }
```

johnthagen:
___
*[src/types/user.tsx, line 5 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCnyT-9DtOXYwudKoj:-MJCnyT-9DtOXYwudKok:be49p1k) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/src/types/user.tsx#L5)):*
> ```XML
> export class User {
>   id: string;
>   avatar: string;
> ```

Here's an example where if we encoded "missing" into the actual type system from the backend all the way to the frontend, we could express that here like:

```ts
avatar?: string;
```

johnthagen (changes requested):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)* status: all files reviewed, 6 unresolved discussions (waiting on @imnasnainaec)

johnthagen:
___
*[src/types/user.tsx, line 3 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCpv3VA--3Vcwt2XJZ:-MJCpv3VA--3Vcwt2XJ_:b-yylays) ([raw file](https://github.com/sillsdev/thecombine/blob/c564efc9e8de7e8bcc62ecf5c1a85d3dea5ddf02/src/types/user.tsx#L3)):*
> ```XML
> import { Hash } from "../goals/MergeDupGoal/MergeDupStep/MergeDupsTree";
> 
> export class User {
> ```

This entire class could be simplified to:

```ts
export class User {
  id: string = "";
  avatar: string = "";
  hasAvatar: boolean = false;
  email: string = "";
  phone: string = "";
  otherConnectionField: string = "";
  projectRoles: Hash<string> = {};
  workedProjects: Hash<string> = {};
  agreement: boolean = false;
  uiLang: string = "";
  token: string = "";
  isAdmin: boolean = false;

  constructor(
    public name: string,
    public username: string,
    public password: string
  ) {}
}
```

This uses [parameter properties](https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties) to remove a large amount of duplication.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)* status: 6 of 8 files reviewed, 7 unresolved discussions (waiting on @imnasnainaec, @jasonleenaylor, and @johnthagen)

___
*[Backend/Models/User.cs, line 19 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJChhN1CTCCWQsdSVQj:-MJE-eVM6NuhC8n62iTm:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/Backend/Models/User.cs#L19)):*

This should be declared as

```csharp
public bool HasAvatar { get; set; }
```

for consistency 
</blockquote></details>

Done.
___
*[Backend/Services/UserApiServices.cs, line 178 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCkuJo5tPdwK7VoyIE:-MJE-fqM2dkrbcBdLOkV:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/Backend/Services/UserApiServices.cs#L178)):*

This can be simplified to:

```c#
users.ForEach(Sanitize);
```
</blockquote></details>

Done.

johnthagen:
___
*[Backend/Models/User.cs, line 19 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCP8r3Db_BjinkcPNT:-MJNkuFc3GAAY4HIbIoZ:b-mf6eob) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/Backend/Models/User.cs#L19)):*

Opened a separate issue about tracking nullables project-wide, as that could be addressed separately from this: https://github.com/sillsdev/TheCombine/issues/754
</blockquote></details>

Done.

Deferring to https://github.com/sillsdev/TheCombine/issues/754

johnthagen:
___
*[src/types/user.tsx, line 5 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCnyT-9DtOXYwudKoj:-MJNl5HeAZ7keZTlfIUG:b8yzd3g) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/src/types/user.tsx#L5)):*

Here's an example where if we encoded "missing" into the actual type system from the backend all the way to the frontend, we could express that here like:

```ts
avatar?: string;
```
</blockquote></details>

Deferred to https://github.com/sillsdev/TheCombine/issues/754

johnthagen (changes requested):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)* status: all files reviewed, 3 unresolved discussions (waiting on @imnasnainaec)

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)* status: 5 of 8 files reviewed, 3 unresolved discussions (waiting on @jasonleenaylor and @johnthagen)

___
*[src/components/ProjectSettings/ProjectUsers/ActiveUsers.tsx, line 39 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCmVRK92n6FrXWSgUQ:-MJSOEeO5LJ409RzdNFj:buxlzoo) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/src/components/ProjectSettings/ProjectUsers/ActiveUsers.tsx#L39)):*

This can be simplified with a `for..of` loop:


```typescript
        this.setState({ projUsers });
        const userAvatar = this.state.userAvatar;
        // for loop rather than .forEach forces each await to finish
        for (let user of projUsers) {
          if (user.hasAvatar) {
            userAvatar[user.id] = await avatarSrc(user.id);
          }
        }
        this.setState({ userAvatar });
```
</blockquote></details>

Use the final method from https://lavrton.com/javascript-loops-how-to-handle-async-await-6252dd3c795/ instead.
___
*[src/components/ProjectSettings/ProjectUsers/ProjectUsers.tsx, line 82 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCneYG5NN_w-pifwpN:-MJSOIYUD3CjtSbgDy-5:buxlzoo) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/src/components/ProjectSettings/ProjectUsers/ProjectUsers.tsx#L82)):*

Can be simplified with `for...of`:

```ts
            const userAvatar = this.state.userAvatar;
            // for loop rather than .forEach forces each await to finish
            for (let user of returnedUsers) {
              if (user.hasAvatar) {
                userAvatar[user.id] = await backend.avatarSrc(user.id);
              }
            }
```
</blockquote></details>

Use the final method from https://lavrton.com/javascript-loops-how-to-handle-async-await-6252dd3c795/ instead.
___
*[src/types/user.tsx, line 3 at r3](https://reviewable.io/reviews/sillsdev/thecombine/742#-MJCpv3VA--3Vcwt2XJZ:-MJSOO_10_vfR7-RXv_a:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ddefffa9e867e1311f42735da0f02262efca46c0/src/types/user.tsx#L3)):*

This entire class could be simplified to:

```ts
export class User {
  id: string = "";
  avatar: string = "";
  hasAvatar: boolean = false;
  email: string = "";
  phone: string = "";
  otherConnectionField: string = "";
  projectRoles: Hash<string> = {};
  workedProjects: Hash<string> = {};
  agreement: boolean = false;
  uiLang: string = "";
  token: string = "";
  isAdmin: boolean = false;

  constructor(
    public name: string,
    public username: string,
    public password: string
  ) {}
}
```

This uses [parameter properties](https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties) to remove a large amount of duplication.
</blockquote></details>

Done.

johnthagen (approved):
> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/742)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #744, opened 2020-09-30
Prune container registry

jmgrady:
This pull request adds a python script, `./docker_deploy/bin/clean_repo.py`, to clean out docker images from the AWS Elastic Container Repository.  The intended use is to automatically remove older images created as part of building for the QA server.  It can also be used manually to manage the released images.  The usage is:
```
$ docker_deploy/bin/clean_repo.py --help
usage: clean-repo.py [-h] [--profile PROFILE] [--keep KEEP [KEEP ...]]
                     [--keep-pattern KEEP_PATTERN [KEEP_PATTERN ...]]
                     [--dry-run] [--verbose]
                     repo

Clean an AWS ECR repository of all images/tags except those listed

positional arguments:
  repo                  Docker image repository to be cleaned.

optional arguments:
  -h, --help            show this help message and exit
  --profile PROFILE     AWS user profile to use to connect to AWS ECR
                        (default: None)
  --keep KEEP [KEEP ...]
                        list of tags to keep (default: None)
  --keep-pattern KEEP_PATTERN [KEEP_PATTERN ...]
                        list of regular expressions that specify tags to keep
                        (default: None)
  --dry-run             print delete command that would be run instead of
                        executing it (default: False)
  --verbose             print intermediate values to aid in debugging
                        (default: False)
```

In addition, this PR includes the following changes:
 * changes to the role `aws_access`:
      - it is now less Combine specific, that is, it no longer assumes that the user being setup is `combine`;
      - added the option to specify a default AWS profile when running the `aws_access` role;
      - the AWS region is specified for each profile in aws.yml
 * create a playbook to setup a TeamCity build agent to build TheCombine.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[docker_deploy/bin/clean-repo.py, line 37 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVbuID95bSB9638sD8:-MIVbsiO7e5Ga-8L1Ssm:b-3u3z9k) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L37)):*
> ```Python
>         "--keep",
>         nargs='+',
>         help="list of tags to keep",
> ```

"list... -> "List... (here and once below)
and "print... -> "Print (twice below)
___
*[docker_deploy/bin/clean-repo.py, line 53 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVbS9RAnZrwFC57CGs:-MIVbS9RAnZrwFC57CGt:by9uc5w) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L53)):*
> ```Python
>     )
>     parser.add_argument(
>         "--verbose",
> ```

Can we add `"-v",` in the line before `"--verbose",` to promote lazy debugging?
___
*[docker_deploy/bin/clean-repo.py, line 85 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVdXBx9Sardm2CVeR4:-MIVdXBy-7utYu0kwPav:b-s92osp) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L85)):*
> ```Python
>     args = parse_args()
> 
>     # get a list of the current image tags for the specified repo
> ```

Comment capitalization is inconsistent in this file
___
*[docker_deploy/bin/clean-repo.py, line 87 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVcRgmAHSfqjlX7jzM:-MIVcRgmAHSfqjlX7jzN:b-tu3bbz) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L87)):*
> ```Python
>     # get a list of the current image tags for the specified repo
>     awsCmd = buildAwsCommand(args.profile, args.repo, "describe-images")
>     awsResult = runAwsCmd(awsCmd, args.verbose)
> ```

Is`runAwsCmd` missing `args.dry_run` as the third argument?
___
*[docker_deploy/bin/clean-repo.py, line 114 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVdt8D4_qmiBBqvI2V:-MIVdt8D4_qmiBBqvI2W:bi5dtlv) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L114)):*
> ```Python
>                 print("Testing tag: ", tag, " from ", imageStruct['imagePushedAt'])
>             # check to see if there are patterns to test
>             if len(keep_pattern) > 0 and not re.match(keep_pattern, tag):
> ```

`if len(keep_pattern) > 0 and ...` could be `if keep_pattern and ...` (same with `oldTags` below) but I'm not sure what our standard is for that.

Similarly the `!= None` is unnecessary in the `if` statements above and below-- in fact, removing it could be a defense against the case that `args.keep_pattern` or `args.keep` is an empty list (though that should be impossible with `nargs='+'`).

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[docker_deploy/bin/clean-repo.py, line 53 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVbS9RAnZrwFC57CGs:-MIYsdna03nLrVIJsW8x:b4ak877) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L53)):*

Can we add `"-v",` in the line before `"--verbose",` to promote lazy debugging?
</blockquote></details>

By default `argparse` will  allow abbreviations if they are unambiguous.  See _ArgumentParser Objects_ in https://docs.python.org/3/library/argparse.html.  This means that `--v` is equivalent to `--verbose`.  Will this work for you?  I would rather not add extra code (which requires maintenance).
___
*[docker_deploy/bin/clean-repo.py, line 87 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVcRgmAHSfqjlX7jzM:-MIYu_UaA3rKih-jNBz8:b-yjmbiu) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L87)):*

Is`runAwsCmd` missing `args.dry_run` as the third argument?
</blockquote></details>

No, it is not in this case.  `args.dry_run` would prevent the `describe-images` command from being run.  This is needed for the rest of the script so that we can generate the command to remove the images/tags that we want.  `args.dry_run` is passed to the second call to print out the command that we would run to delete the images/tags without actually deleting them.
___
*[docker_deploy/bin/clean-repo.py, line 114 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVdt8D4_qmiBBqvI2V:-MIYxkJY3NLYkKMX260t:bfisnzv) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L114)):*

`if len(keep_pattern) > 0 and ...` could be `if keep_pattern and ...` (same with `oldTags` below) but I'm not sure what our standard is for that.

Similarly the `!= None` is unnecessary in the `if` statements above and below-- in fact, removing it could be a defense against the case that `args.keep_pattern` or `args.keep` is an empty list (though that should be impossible with `nargs='+'`).
</blockquote></details>

Done.

Actually, while testing this improvement I found a related bug.  The logic for `args.keep` was wrong.
```
if args.keep != None and tag not in args.keep:
```
is now corrected to:
```
if not args.keep or tag not in args.keep:
```
(line 116)

The way I had written it is if I had not specified any specific tags to keep, it would keep them all instead of keeping none of the tags that did not match the keep-pattern.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[docker_deploy/bin/clean-repo.py, line 53 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVbS9RAnZrwFC57CGs:-MIZU31gAHYaC12w2E22:b56e3ay) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L53)):*

By default `argparse` will  allow abbreviations if they are unambiguous.  See _ArgumentParser Objects_ in https://docs.python.org/3/library/argparse.html.  This means that `--v` is equivalent to `--verbose`.  Will this work for you?  I would rather not add extra code (which requires maintenance).
</blockquote></details>

Oh, that's cool! Yeah, it's good as is.
___
*[docker_deploy/bin/clean-repo.py, line 87 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVcRgmAHSfqjlX7jzM:-MIZUOV3C5mUMlowHsU4:b-hjh41y) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L87)):*

No, it is not in this case.  `args.dry_run` would prevent the `describe-images` command from being run.  This is needed for the rest of the script so that we can generate the command to remove the images/tags that we want.  `args.dry_run` is passed to the second call to print out the command that we would run to delete the images/tags without actually deleting them.
</blockquote></details>

Oh, I didn't see any purpose for `dry_run` but now I see it's used at the bottom.

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 85 at r1](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIVdXBx9Sardm2CVeR4:-MIdAfktCs3ypvcY85Ix:b224y70) ([raw file](https://github.com/sillsdev/thecombine/blob/869a6eb31a65cd1bced559b6b5f4288e0749f373/docker_deploy/bin/clean-repo.py#L85)):*

# Comment capitalization is inconsistent in this file
</blockquote></details>

Whoa, big font there! ;)

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdAlQS1yOhSPTZA_E1:-MIdAlQS1yOhSPTZA_E2:bv8ftn0) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*
> ```Python
>     return parser.parse_args()
> 
> def runAwsCmd(awsCmd, verbose=False, dry_run=False):
> ```

We should follow the standard PEP8 style guide for Python code, which prescribes snake_case (`run_aws_cmd`).

I'll create another PR where we add some Python linting so that this kind of thing is automatically picked up.

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 1 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdBgVp2_r6MUWpvyWv:-MIdBgVq7zBO6kwvyeKD:b-1rjhqj) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L1)):*
> ```Python
> #!/usr/bin/env python3
> ```

This file should be renamed to `clean_repo.py` to follow standard Python conventions (it's named that in the PR first comment, but not in the file).

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 14 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdBw1M1sW4tPSLnDw1:-MIdBw1M1sW4tPSLnDw2:b-tm5cx1) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L14)):*
> ```Python
> import argparse
> import json
> import os
> ```

`os` is an unused import.

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 102 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdC4ea-SPm_jt8lCuD:-MIdC4ea-SPm_jt8lCuE:b-x84uj1) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L102)):*
> ```Python
>     # Join patterns of tags to keep to a single regular expression
>     keep_pattern = ''
>     if args.keep_pattern != None:
> ```

The idiomatic way to compare to `None` is Python is using the identity operator `is`.

`if args.keep_pattern is not None:`

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 74 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCImfCWJb0WHVEBCV:-MIdCImfCWJb0WHVEBCW:bk7qps3) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L74)):*
> ```Python
> 
> def buildAwsCommand(profile, repo, subcommand, awsArgs=None):
>     awsCmd = [ "aws", "ecr" ]
> ```

All local variables in Python should be in `snake_case` (`awd_cmd`).

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 125 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCpRpDMExRdcNbGB6:-MIdCpRpDMExRdcNbGB7:bku27kj) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L125)):*
> ```Python
>             image_ids.append("imageTag="+tag)
>         awsCmd = buildAwsCommand(args.profile, args.repo, "batch-delete-image", image_ids)
>         awsResult = runAwsCmd(awsCmd, args.verbose, args.dry_run)
> ```

`awsResult` is unused, does anything need to be checked with this? Or should we just not assign the output to a variable?

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 128 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCytxBoq3KmK0E6kV:-MIdCytxBoq3KmK0E6kW:b-lj9200) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L128)):*
> ```Python
>     elif args.verbose:
>         print("No images/tags were deleted.")
> # Standard boilerplate to call main().
> ```

We can probably remove this comment as those familiar with Python enough to read much of anything in this file will be familiar with how `main` is called in Python.

johnthagen:
@jmgrady Is there any way for me to test this locally?

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIdPueK8dptgH_2TPTg:b-uyihfs) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*
> ```Python
>         return awsResults
> 
> def buildAwsCommand(profile, repo, subcommand, awsArgs=None):
> ```

We should add type hints to this code to improve static analysis and readability

```py
from typing import List, Optional

def build_aws_profile(profile: Optional[str], repo: List[str], subcommand: str, awsArgs: Optional[str] = None):
```

In another PR I'll add automated type checking.

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdQZj_AEXa8a5s7nGw:-MIdQZj_AEXa8a5s7nGx:b-cvennm) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*
> ```Python
>     return parser.parse_args()
> 
> def runAwsCmd(awsCmd, verbose=False, dry_run=False):
> ```

Type hints:

```py
def run_aws_cmd(aws_cmd: str, verbose: Optional[bool] = False, dry_run: Optional[bool] = False):
```

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIdQxlr1ifpdgIGkmMk:b81dkzc) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*
> from typing import List, Optional
> 
> def build_aws_profile(profile: Optional[str], repo: List[str], subcommand: str, awsArgs: Optional[str] = None):

Fixed with return type:

```py
from typing import List, Optional

def build_aws_profile(profile: Optional[str], repo: List[str], subcommand: str, awsArgs: Optional[str] = None) -> List[str]:
```

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdQZj_AEXa8a5s7nGw:-MIdR9uS3mUC4gtinoKe:b-ujvbdu) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*
> def run_aws_cmd(aws_cmd: str, verbose: Optional[bool] = False, dry_run: Optional[bool] = False):

Fixed with return type:

```py
def run_aws_cmd(aws_cmd: str, verbose: Optional[bool] = False, dry_run: Optional[bool] = False) -> subprocess.CompletedProcess:
```

johnthagen (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, 9 unresolved discussions (waiting on @jmgrady and @johnthagen)

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, 9 unresolved discussions (waiting on @johnthagen)

___
*[docker_deploy/bin/clean-repo.py, line 1 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdBgVp2_r6MUWpvyWv:-MIdWbuX0oU0xObHn_0N:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L1)):*

This file should be renamed to `clean_repo.py` to follow standard Python conventions (it's named that in the PR first comment, but not in the file).
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean-repo.py, line 14 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdBw1M1sW4tPSLnDw1:-MIdPvU_DPNr49orRoYC:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L14)):*

`os` is an unused import.
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdAlQS1yOhSPTZA_E1:-MIdQAuP9FO9V1h9FnyF:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*

We should follow the standard PEP8 style guide for Python code, which prescribes snake_case (`run_aws_cmd`).

I'll create another PR where we add some Python linting so that this kind of thing is automatically picked up.
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdQZj_AEXa8a5s7nGw:-MIdQwH9CD8gGBDZ1y8s:b-j5rooo) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*

Type hints:

```py
def run_aws_cmd(aws_cmd: str, verbose: Optional[bool] = False, dry_run: Optional[bool] = False):
```
</blockquote></details>

What is the value of "Optional"?  If I define a default value it is by nature an optional argument.
___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIdRvlG-RHTgZXcV9mT:b-sr6y75) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*

> from typing import List, Optional
> 
> def build_aws_profile(profile: Optional[str], repo: List[str], subcommand: str, awsArgs: Optional[str] = None):

Fixed with return type:

```py
from typing import List, Optional

def build_aws_profile(profile: Optional[str], repo: List[str], subcommand: str, awsArgs: Optional[str] = None) -> List[str]:
```
</blockquote></details>

In this case, profile is not optional, the argument must be present but it may be "None".  Also, it will be a list since, if I understand correctly, that is how argparse will return it.  Perhaps I don't understand the meaning of optional in this case.
___
*[docker_deploy/bin/clean-repo.py, line 74 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCImfCWJb0WHVEBCV:-MIdTkrE90yd6k6WAdcR:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L74)):*

All local variables in Python should be in `snake_case` (`awd_cmd`).
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean-repo.py, line 102 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdC4ea-SPm_jt8lCuD:-MIdTv2Y1wXJKU76R5iY:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L102)):*

The idiomatic way to compare to `None` is Python is using the identity operator `is`.

`if args.keep_pattern is not None:`
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean-repo.py, line 125 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCpRpDMExRdcNbGB6:-MIdU4kq5TH725klNtu9:b5ln5e) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L125)):*

`awsResult` is unused, does anything need to be checked with this? Or should we just not assign the output to a variable?
</blockquote></details>

Good point.  Earlier, I was printing stdout and stderr if args.verbose was True.  I think it will be helpful to have a list of images/tags deleted in the TeamCity build logs so I will print aws_result.stdout (on STDOUT) & aws_result.stderr (on STDERR) independent of args.verbose.
___
*[docker_deploy/bin/clean-repo.py, line 128 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCytxBoq3KmK0E6kV:-MIdW9ie155UXIdSnZBd:bwzh3rv) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L128)):*

We can probably remove this comment as those familiar with Python enough to read much of anything in this file will be familiar with how `main` is called in Python.
</blockquote></details>

It's for _my_ benefit! :-)  Done.

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdQZj_AEXa8a5s7nGw:-MIdanah1Kzh7TYAGX-L:b-opoi3d) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*

What is the value of "Optional"?  If I define a default value it is by nature an optional argument.
</blockquote></details>

Please ignore `Optional` on this. My mistake.

```py
def run_aws_cmd(aws_cmd: str, verbose: bool = False, dry_run: bool = False) -> subprocess.CompletedProcess:
```

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 128 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCytxBoq3KmK0E6kV:-MIdbNgY9ijg5vxTf5Hi:b32vhw) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L128)):*

It's for _my_ benefit! :-)  Done.
</blockquote></details>

:)

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 32 at r3](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIddYzD2ltn_XkGlIgz:-MIddYzD2ltn_XkGlIh-:bcfxe40) ([raw file](https://github.com/sillsdev/thecombine/blob/bec47f42e8dadbbbffd5d5842c48d3f970fd3755/docker_deploy/bin/clean_repo.py#L32)):*
> ```Python
>     parser.add_argument(
>         "repo",
>         nargs=1,
> ```

We should remove `nargs` if only one repo can be passed.  That will bring repo in as just a `str`.

Then we can change the type hint of `build_aws_cmd(..., repo: str, ...)`

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIdeFD37syDjzoXoi4S:b-efj13h) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*

In this case, profile is not optional, the argument must be present but it may be "None".  Also, it will be a list since, if I understand correctly, that is how argparse will return it.  Perhaps I don't understand the meaning of optional in this case.
</blockquote></details>

Per conversation, `profile: Optional[str]` just means that `profile` must be passed, but from a type system perspective, what is passed can either be a `str` or `None`.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 27 at r3](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdeatS9syeDktGp3Dn:-MIdeatS9syeDktGp3Do:b-21fuu1) ([raw file](https://github.com/sillsdev/thecombine/blob/bec47f42e8dadbbbffd5d5842c48d3f970fd3755/docker_deploy/bin/clean_repo.py#L27)):*
> ```Python
>     parser.add_argument(
>         "--profile",
>         nargs=1,
> ```

Should this `nargs` be removed? In general you don't need to use `nargs` if you only want `argparse` to limit the user to one input for that arg. That will get rid of the extra `list` that it's wrapped in, I think.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 77 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdjTrP5OUJe516ldOF:-MIdjTrP5OUJe516ldOG:b59xkfd) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L77)):*
> ```Python
>     aws_cmd = [ "aws", "ecr" ]
>     if profile:
>         aws_cmd.append("--profile="+profile[0])
> ```

Now that `profile` is a `str`, should the `[0]` be removed?

Also, we can use format strings to make this look a little nicer:

```py
aws_cmd.append(f"--profile={profile}")
```

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 78 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdjnhND6VICfbKK3cf:-MIdjnhND6VICfbKK3cg:bs3zu2h) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L78)):*
> ```Python
>     if profile:
>         aws_cmd.append("--profile="+profile[0])
>     aws_cmd.extend(["--repository-name="+repo[0], "--output=json", subcommand])
> ```

Similar comment here:

```py
 aws_cmd.extend([f"--repository-name={repo}", "--output=json", subcommand])
```

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdQZj_AEXa8a5s7nGw:-MIdk4t85NuBuABuD_kl:b-dcs0md) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*
> def run_aws_cmd(aws_cmd: str, verbose: bool = False, dry_run: bool = False) -> subprocess.CompletedProcess:

I see that `aws_cmd` should be a `list`. We also need `Optional` on the output because `dry_run` causes the function to return `None`:

```py
def run_aws_cmd(aws_cmd: List[str], verbose: bool = False, dry_run: bool = False) -> Optional[subprocess.CompletedProcess]:
```

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 128 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdkpQlEQjVkMyONdid:-MIdkpQlEQjVkMyONdie:b-5wfnmp) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L128)):*
> ```Python
>         aws_result = run_aws_cmd(aws_cmd, args.verbose, args.dry_run)
>         print(aws_result.stdout)
>         print(aws_result.stderr, file=sys.stderr)
> ```

`sys.stderr` -> `stderr`, or perhaps better, change to `import sys` at  top of file.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 122 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlBk2EsIcGoRzuR7B:-MIdlBk2EsIcGoRzuR7C:bk4cijs) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L122)):*
> ```Python
>     # Remove all the specified image(s)
>     if old_tags:
>     # Convert the list of tags to a set of image-ids for the AWS ECR command
> ```

Minor, but this comment should be indented to the same amount as the source below it. My code-formatting PR for Python will probably automatically fix all these kinds of things moving forward.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 113 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlSRu325HYIwhDg6U:-MIdlSRu325HYIwhDg6V:b-59sjh7) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L113)):*
> ```Python
>         for tag in image_struct['imageTags']:
>             if args.verbose:
>                 print("Testing tag: ", tag, " from ", image_struct['imagePushedAt'])
> ```

Use format string

```py
print(f"Testing tag: {tag} from {image_struct['imagePushedAt']}")
```

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 124 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlgLf1ZSUv0_4IFDo:-MIdlgLf1ZSUv0_4IFDp:b-1jmz3w) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L124)):*
> ```Python
>     # Convert the list of tags to a set of image-ids for the AWS ECR command
>         for tag in old_tags:
>             image_ids.append("imageTag="+tag)
> ```

`image_ids.append(f"imageTag={tag}")`

johnthagen (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, 11 unresolved discussions (waiting on @jmgrady)

___
*[docker_deploy/bin/clean_repo.py, line 125 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlfbR4FdlouWz1eiJ:-MIdlfbR4FdlouWz1eiK:b-xqd5d1) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L125)):*
> ```
>  image_ids.append("imageTag="+tag)
> ```

jmgrady (commented):
I will add instructions for testing locally in Slack.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744#-:-MIdi4Q82Hjp6kkLgk1_:b-ooafi8)* status: 0 of 10 files reviewed, 10 unresolved discussions (waiting on @jmgrady and @johnthagen)

___
*[docker_deploy/bin/clean-repo.py, line 59 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdQZj_AEXa8a5s7nGw:-MIdjfZIA-Y7dn1r3Pd5:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L59)):*

Please ignore `Optional` on this. My mistake.

```py
def run_aws_cmd(aws_cmd: str, verbose: bool = False, dry_run: bool = False) -> subprocess.CompletedProcess:
```
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIdji73Bi0_pv7au0lI:bwayu87) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*

Per conversation, `profile: Optional[str]` just means that `profile` must be passed, but from a type system perspective, what is passed can either be a `str` or `None`.
</blockquote></details>

The specification above shows awsArgs (now renamed to aws_args) as `Optional[str]` but it is really an optional list of strings.  How do we specify that?  `Optional[List[str]]`?
___
*[docker_deploy/bin/clean-repo.py, line 125 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdCpRpDMExRdcNbGB6:-MIdlXwJ0OR8xQrgkcJu:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L125)):*

Good point.  Earlier, I was printing stdout and stderr if args.verbose was True.  I think it will be helpful to have a list of images/tags deleted in the TeamCity build logs so I will print aws_result.stdout (on STDOUT) & aws_result.stderr (on STDERR) independent of args.verbose.
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 27 at r3](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdeatS9syeDktGp3Dn:-MIdikjcEWkwsFxA9PVT:bjw1tzy) ([raw file](https://github.com/sillsdev/thecombine/blob/bec47f42e8dadbbbffd5d5842c48d3f970fd3755/docker_deploy/bin/clean_repo.py#L27)):*

Should this `nargs` be removed? In general you don't need to use `nargs` if you only want `argparse` to limit the user to one input for that arg. That will get rid of the extra `list` that it's wrapped in, I think.
</blockquote></details>

Yes, that does the trick.  Done.
___
*[docker_deploy/bin/clean_repo.py, line 32 at r3](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIddYzD2ltn_XkGlIgz:-MIdjQfx9DgpmeC5LO8z:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/bec47f42e8dadbbbffd5d5842c48d3f970fd3755/docker_deploy/bin/clean_repo.py#L32)):*

We should remove `nargs` if only one repo can be passed.  That will bring repo in as just a `str`.

Then we can change the type hint of `build_aws_cmd(..., repo: str, ...)`
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 77 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdjTrP5OUJe516ldOF:-MIdlAtX4QyvBL0ZtDX1:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L77)):*

Now that `profile` is a `str`, should the `[0]` be removed?

Also, we can use format strings to make this look a little nicer:

```py
aws_cmd.append(f"--profile={profile}")
```
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 78 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdjnhND6VICfbKK3cf:-MIdlBEkBpM1gLprV5hu:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L78)):*

Similar comment here:

```py
 aws_cmd.extend([f"--repository-name={repo}", "--output=json", subcommand])
```
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 124 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlgLf1ZSUv0_4IFDo:-MIdmbbIB_DX24elrC_W:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L124)):*

`image_ids.append(f"imageTag={tag}")`
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 125 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlfbR4FdlouWz1eiJ:-MIdmgo8-wM8acuYX5Q0:b-6a05lx) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L125)):*

> ```
>  image_ids.append("imageTag="+tag)
> ```

</blockquote></details>

??  I suspect this was an extra click since there is the comment above (and it's "Done").
___
*[docker_deploy/bin/clean_repo.py, line 128 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdkpQlEQjVkMyONdid:-MIdlYHf60nj2IeIbtLM:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L128)):*

`sys.stderr` -> `stderr`, or perhaps better, change to `import sys` at  top of file.
</blockquote></details>

Done.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 128 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdkpQlEQjVkMyONdid:-MIdoOhA3_ghsa5U_eKY:b-8vkwst) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L128)):*

Done.
</blockquote></details>

Sorry, if both of the fixes are applied, it won't be referenced.

Should be changed here to `sys.stderr` and leave the import as is above.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 125 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlfbR4FdlouWz1eiJ:-MIdo_Or5H_RMJhGSMEo:b-1ur390) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L125)):*

??  I suspect this was an extra click since there is the comment above (and it's "Done").
</blockquote></details>

LGTM.

No idea what happened there.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 124 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlgLf1ZSUv0_4IFDo:-MIdoipx8v3p6n5aq3Gv:b13kptf) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L124)):*

Done.
</blockquote></details>

This change doesn't look like it got committed?

johnthagen:
___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIdzDlVAp2kbBv33yDq:b-l7c5ut) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*

The specification above shows awsArgs (now renamed to aws_args) as `Optional[str]` but it is really an optional list of strings.  How do we specify that?  `Optional[List[str]]`?
</blockquote></details>

Yes, `Optional[List[str]]`

johnthagen (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @jmgrady)

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @jmgrady and @johnthagen)

___
*[docker_deploy/bin/clean-repo.py, line 73 at r2](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdPueK8dptgH_2TPTf:-MIeNnXJ3CXdc36mOJS2:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/ff3be4002f20e457c9a003520430a507f787d146/docker_deploy/bin/clean-repo.py#L73)):*

Yes, `Optional[List[str]]`
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 113 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlSRu325HYIwhDg6U:-MIeNy5X6GIuJvMbtzTw:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L113)):*

Use format string

```py
print(f"Testing tag: {tag} from {image_struct['imagePushedAt']}")
```

</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 125 at r4](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIdlfbR4FdlouWz1eiJ:-MIdnMHH2XQp2hmxVPtv:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cb890b45bc492f63f559c92775f2e7e4129ce7e8/docker_deploy/bin/clean_repo.py#L125)):*

??  I suspect this was an extra click since there is the comment above (and it's "Done").
</blockquote></details>

Done.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 39 at r5](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeQwxrEQEeRjagE36c:-MIeQwxrEQEeRjagE36d:b-h0vec5) ([raw file](https://github.com/sillsdev/thecombine/blob/fda2dc68b5c1a84d39a322923a01bd9960e118a7/docker_deploy/bin/clean_repo.py#L39)):*
> ```Python
>     )
>     parser.add_argument(
>         "--keep-pattern",
> ```

Given this tool can delete data, should we default the keep pattern to SemVer release pattern? (E.g. `"\d+\.\d+\.\d+"`)?

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 27 at r5](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeRJc68w65Cryub99q:-MIeRJc68w65Cryub99r:b3oqqa4) ([raw file](https://github.com/sillsdev/thecombine/blob/fda2dc68b5c1a84d39a322923a01bd9960e118a7/docker_deploy/bin/clean_repo.py#L27)):*
> ```Python
>     parser.add_argument(
>         "--profile",
>         help="AWS user profile to use to connect to AWS ECR"
> ```

Is there a finite list of user profiles that can be used? If so, we could consider adding `choices=[...]` argument to enforce the user types in the correct profile.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 7 at r5](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeSkxlEFmh5wewdQ0V:-MIeSkxlEFmh5wewdQ0W:b-to4r6m) ([raw file](https://github.com/sillsdev/thecombine/blob/fda2dc68b5c1a84d39a322923a01bd9960e118a7/docker_deploy/bin/clean_repo.py#L7)):*
> ```Python
> 
> Assumptions:
>   1. aws-cli is installed (see https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html)
> ```

Could we specify which AWS CLI (1 or 2) should be used?

scripts/printVersion.js
```diff
@@ -1 +1 @@
-console.log(require('./package.json').version);
```
johnthagen, line 1: Good catch.

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 18 files reviewed, 3 unresolved discussions (waiting on @jmgrady and @johnthagen)

___
*[docker_deploy/bin/clean_repo.py, line 7 at r5](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeSkxlEFmh5wewdQ0V:-MIeXyDs15ekLzmCZ35e:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/fda2dc68b5c1a84d39a322923a01bd9960e118a7/docker_deploy/bin/clean_repo.py#L7)):*

Could we specify which AWS CLI (1 or 2) should be used?
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 27 at r5](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeRJc68w65Cryub99q:-MIeY-ZX0BgQ9-aQj0rR:b-dgb3al) ([raw file](https://github.com/sillsdev/thecombine/blob/fda2dc68b5c1a84d39a322923a01bd9960e118a7/docker_deploy/bin/clean_repo.py#L27)):*

Is there a finite list of user profiles that can be used? If so, we could consider adding `choices=[...]` argument to enforce the user types in the correct profile.
</blockquote></details>

No, there is no set list.
___
*[docker_deploy/bin/clean_repo.py, line 39 at r5](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeQwxrEQEeRjagE36c:-MIeY9uHApuhriQuTr_M:bec8ct5) ([raw file](https://github.com/sillsdev/thecombine/blob/fda2dc68b5c1a84d39a322923a01bd9960e118a7/docker_deploy/bin/clean_repo.py#L39)):*

Given this tool can delete data, should we default the keep pattern to SemVer release pattern? (E.g. `"\d+\.\d+\.\d+"`)?
</blockquote></details>

I don't think that would be what we want.  Normally we use the keep_pattern to not delete any released versions when we are cleaning up old QA versions - and that is the original use envisioned for the tool.  However, we may also need it to run it manually to delete old releases.  If by default we always keep old releases then we have to provide a bogus --keep-pattern to override the default.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 8 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeZcbZ--qMnbRAF2n7:-MIeZcbZ--qMnbRAF2n8:b-9gial3) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L8)):*
> ```Python
> Assumptions:
>   1. aws-cli version 2 is installed (see https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html)
>   2. aws-cli has been configured with the access keys required to read/write to the
> ```

Could we add specifically what files need to be added (for a user testing the script): `~/.aws/....`

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 119 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIe_70xBpoW0wYIy55V:-MIe_70y6ad0UvCZRdUK:b-dta19j) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L119)):*
> ```Python
> 
>     # Remove all the specified image(s)
>     if old_tags:
> ```

Similar to how we explicitly check for `if __ is not None`,  I think this could be a little more readable if we check for:

```py
if len(old_tags) > 0:
```

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 0 of 18 files reviewed, 2 unresolved discussions (waiting on @jmgrady and @johnthagen)

___
*[docker_deploy/bin/clean_repo.py, line 8 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeZcbZ--qMnbRAF2n7:-MIe_K5v1Zw8uC3h2rJg:b39m650) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L8)):*

Could we add specifically what files need to be added (for a user testing the script): `~/.aws/....`
</blockquote></details>

Since I don't really want to put details such as access keys into the script, I left it general.  I can add a comment to suggest that one way to do that is to run `aws configure`
___
*[docker_deploy/bin/clean_repo.py, line 119 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIe_70xBpoW0wYIy55V:-MIea-C3Ame-WPczo8CH:bx2axwz) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L119)):*

Similar to how we explicitly check for `if __ is not None`,  I think this could be a little more readable if we check for:

```py
if len(old_tags) > 0:
```
</blockquote></details>

:-)
That's what I had originally.  Danny suggested that I change it to the simpler form of `if old_tags` to handle `None` or an empty list.  In this case, it cannot be `None` since we set it to an empty list on line 89 but this would make it consistent with checking `keep_pattern` earlier.  Your call.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 119 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIe_70xBpoW0wYIy55V:-MIebDy99bQmCNAd_0Oh:bwklcl3) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L119)):*

:-)
That's what I had originally.  Danny suggested that I change it to the simpler form of `if old_tags` to handle `None` or an empty list.  In this case, it cannot be `None` since we set it to an empty list on line 89 but this would make it consistent with checking `keep_pattern` earlier.  Your call.
</blockquote></details>

I personally think it's better to have the explicit check. Since Python is a dynamic language, if there is an accidental assignment of `False`, `0`, or `None` to `old_tags`, this will fail immediately rather than silently being propagate.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 119 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIe_70xBpoW0wYIy55V:-MIebV4O0tGtRjWgu-pc:bwtqtj2) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L119)):*

I personally think it's better to have the explicit check. Since Python is a dynamic language, if there is an accidental assignment of `False`, `0`, or `None` to `old_tags`, this will fail immediately rather than silently being propagate.
</blockquote></details>

Since in this case, `old_tags` can't be set to `None`.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 93 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIe__TL0b4ocTVVjlfi:-MIe__TL0b4ocTVVjlfj:b-3cloyh) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L93)):*
> ```Python
>     # Load the JSON output of the describe-images command into a 'repo_images'
>     # dictionary
>     repo_images = json.loads(aws_result.stdout)
> ```

We can type hint this variable to improve static analysis:

```py
from typing import List, Optional, Dict, Union

AwsJsonResult = Dict[str, List[Dict[str, Union[str, List[str]]]]]

...
    repo_images: AwsJsonResult = json.loads(aws_result.stdout)
```

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 126 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecUf7DsvPcWrlT4Tl:-MIecUf7DsvPcWrlT4Tm:b-6zjp3q) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L126)):*
> ```Python
>         aws_cmd = build_aws_cmd(args.profile, args.repo, "batch-delete-image", image_ids)
>         aws_result = run_aws_cmd(aws_cmd, args.verbose, args.dry_run)
>         print(aws_result.stdout)
> ```

Need to wrap these prints in `if aws_result is not None:` to handle the `--dry-run` case.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 98 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecjNh22ijgn2rb35h:-MIecjNh22ijgn2rb35i:b-sjtnnu) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L98)):*
> ```Python
>     # Initialize list of images to be removed with the option name for the
>     # aws ecr command
>     image_ids = ["--image-ids"]
> ```

To limit the scope of this variable, it could be declare down under `if old_tags:`.

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 90 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecxXRB2TOhmpLEvBt:-MIecxXRB2TOhmpLEvBu:byjw17a) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L90)):*
> ```Python
> 
>     # Create a list of tags that are not on our list of tags to keep
>     old_tags = []
> ```

Should provide a concrete type for this list. When the type checker is applied, this can catch accidently appending other types (like `int`) to the list.

```py
old_tags: List[str] = []
```

The scope of the variable could also be reduced by moving the declaration to just before

```py
for image_struct in repo_images['imageDetails']:
```

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 8 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIeZcbZ--qMnbRAF2n7:-MIeeDW98VpvuVjpQLkX:brxp9sq) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L8)):*

Since I don't really want to put details such as access keys into the script, I left it general.  I can add a comment to suggest that one way to do that is to run `aws configure`
</blockquote></details>

Sounds good.

johnthagen (commented):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 1 of 18 files reviewed, 6 unresolved discussions (waiting on @jmgrady)

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 119 at r7](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIe_70xBpoW0wYIy55V:-MIegEmi0O3HU0xiWmXO:bzge0sq) ([raw file](https://github.com/sillsdev/thecombine/blob/d51133a22bddfc1b930333bd9c64251f2443ff76/docker_deploy/bin/clean_repo.py#L119)):*

Since in this case, `old_tags` can't be set to `None`.
</blockquote></details>

This one is a bit of a style/subjective, so I'll just resolve it. I personally like the explicit style for readability and catching a bug here or there, but many people prefer this style too.

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 1 of 18 files reviewed, 3 unresolved discussions (waiting on @jmgrady and @johnthagen)

___
*[docker_deploy/bin/clean_repo.py, line 90 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecxXRB2TOhmpLEvBt:-MIifDBq3OzaubP1OBDC:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L90)):*

Should provide a concrete type for this list. When the type checker is applied, this can catch accidently appending other types (like `int`) to the list.

```py
old_tags: List[str] = []
```

The scope of the variable could also be reduced by moving the declaration to just before

```py
for image_struct in repo_images['imageDetails']:
```
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 98 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecjNh22ijgn2rb35h:-MIiaiwd4moJCMYG7q_4:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L98)):*

To limit the scope of this variable, it could be declare down under `if old_tags:`.
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 126 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecUf7DsvPcWrlT4Tl:-MIibfNeCN7yWItfpFF9:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L126)):*

Need to wrap these prints in `if aws_result is not None:` to handle the `--dry-run` case.
</blockquote></details>

Done.

johnthagen:
*[a discussion](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIk0g9U5k_0kUTAJTjy:-MIk0g9V7jDQfYMsVMne:b-eldprl) (no related file):*
 Should the `.ts` files be marked executable? They aren't run directly, just used as input to `tsc` which compiles them. I think only the `.bat` file would be executed directly, but only on Windows where that bit isn't used?

---

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 90 at r8](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIecxXRB2TOhmpLEvBt:-MIk1YmtCpgo8xNorZ3H:b6oqdlv) ([raw file](https://github.com/sillsdev/thecombine/blob/13a286d9cd580e7a00ab938b3ab6123de41969bc/docker_deploy/bin/clean_repo.py#L90)):*

Done.
</blockquote></details>

Did this change get checked in?

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 126 at r10](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIk1nIp0CfvZbLvn8Q7:-MIk1nIp0CfvZbLvn8Q8:buzkflo) ([raw file](https://github.com/sillsdev/thecombine/blob/b33042a748a3ce907f5d75a9bc4696a40e382767/docker_deploy/bin/clean_repo.py#L126)):*
> ```Python
>             print(f"AWS Command: {aws_cmd}")
>         else:
>             aws_result = run_aws_cmd(aws_cmd, args.verbose)
> ```

`aws_result` is unused. Should it be logged or inspected?

johnthagen:
___
*[docker_deploy/bin/clean_repo.py, line 59 at r10](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIk1zdrBOxsEy7xfP6L:-MIk1zdrBOxsEy7xfP6M:b-drn85a) ([raw file](https://github.com/sillsdev/thecombine/blob/b33042a748a3ce907f5d75a9bc4696a40e382767/docker_deploy/bin/clean_repo.py#L59)):*
> ```Python
> 
>     aws_results = subprocess.run(
>         aws_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True
> ```

I think we should consider adding `check=True` to the `subprocess.run` command. This will automatically check the return code and throw an exception if it fails. This will have the script fail fast if there is an issue with the `aws` command, rather than have it continue on silently. It's a really simple way to check the return code and fail if it's a failure case (which, we probably can't handle anyway because it's likely something like `aws` isn't installed, or there is no network connection).

https://docs.python.org/3/library/subprocess.html#subprocess.run

johnthagen (commented):
> Reviewed 8 of 10 files at r1, 7 of 8 files at r6, 1 of 3 files at r9, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: all files reviewed, 3 unresolved discussions (waiting on @jmgrady)

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: all files reviewed, 2 unresolved discussions (waiting on @johnthagen)

___
*[docker_deploy/bin/clean_repo.py, line 59 at r10](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIk1zdrBOxsEy7xfP6L:-MIk4PzI1NsZ8kHMPOZS:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/b33042a748a3ce907f5d75a9bc4696a40e382767/docker_deploy/bin/clean_repo.py#L59)):*

I think we should consider adding `check=True` to the `subprocess.run` command. This will automatically check the return code and throw an exception if it fails. This will have the script fail fast if there is an issue with the `aws` command, rather than have it continue on silently. It's a really simple way to check the return code and fail if it's a failure case (which, we probably can't handle anyway because it's likely something like `aws` isn't installed, or there is no network connection).

https://docs.python.org/3/library/subprocess.html#subprocess.run
</blockquote></details>

Done.
___
*[docker_deploy/bin/clean_repo.py, line 126 at r10](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIk1nIp0CfvZbLvn8Q7:-MIk4a368n4fk9cvYDCq:b16nnxb) ([raw file](https://github.com/sillsdev/thecombine/blob/b33042a748a3ce907f5d75a9bc4696a40e382767/docker_deploy/bin/clean_repo.py#L126)):*

`aws_result` is unused. Should it be logged or inspected?
</blockquote></details>

Originally I was printing stdout and stderr always but then I realized that if I use the --verbose flag it will print them.  When we run under TeamCity, we don't get stdout in the build log anyway so I will just remove the assignment to aws_result.

johnthagen:
___
*[scripts/dropDB.ts, line 1 at r12](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIok3fw-AhZgQgazk8o:-MIok3fw-AhZgQgazk8p:b-7eczhz) ([raw file](https://github.com/sillsdev/thecombine/blob/93ba28769428f894d934b4e1327125bc67c5adfc/scripts/dropDB.ts#L1)):*
> ```XML
> import { spawnSync } from "child_process";
> ```

We should remove the executable bit from all of the `scripts/` files that have executable bit set in this PR, since they are not executable directly (other than the `.bat` file, but  that is Windows-only which doesn't use file permissions like this.

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @johnthagen)

___
*[scripts/dropDB.ts, line 1 at r12](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIok3fw-AhZgQgazk8o:-MIsDOOa8f4d2TDQWO3f:b-58dqa9) ([raw file](https://github.com/sillsdev/thecombine/blob/93ba28769428f894d934b4e1327125bc67c5adfc/scripts/dropDB.ts#L1)):*

We should remove the executable bit from all of the `scripts/` files that have executable bit set in this PR, since they are not executable directly (other than the `.bat` file, but  that is Windows-only which doesn't use file permissions like this.
</blockquote></details>

Sort of.  printVersion.js still needs to be executable.  The execute bit has been removed from the *.ts files.

Also, I think that `gen_test_cert.bat` provides no value and should be removed:
  1. it is not cross-platform;
  2. it only consolidates two commands into one; and
  3. the README.md instructs the reader to run the two commands rather than the script.

johnthagen:
___
*[scripts/dropDB.ts, line 1 at r12](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIok3fw-AhZgQgazk8o:-MIsS8p04unllCneuOQf:bwy1w4c) ([raw file](https://github.com/sillsdev/thecombine/blob/93ba28769428f894d934b4e1327125bc67c5adfc/scripts/dropDB.ts#L1)):*

Sort of.  printVersion.js still needs to be executable.  The execute bit has been removed from the *.ts files.

Also, I think that `gen_test_cert.bat` provides no value and should be removed:
  1. it is not cross-platform;
  2. it only consolidates two commands into one; and
  3. the README.md instructs the reader to run the two commands rather than the script.
</blockquote></details>

Yes, let's remove `gen_test_cert.bat`.

Interesting, I assumed running `npm run version` wouldn't need `printVersion.js` to be executable because the file is being read by `node`, not executed directly, but I haven't tested on Linux yet.

johnthagen (commented):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: all files reviewed, 1 unresolved discussion (waiting on @jmgrady)

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: all files reviewed, 1 unresolved discussion (waiting on @johnthagen)

___
*[scripts/dropDB.ts, line 1 at r12](https://reviewable.io/reviews/sillsdev/thecombine/744#-MIok3fw-AhZgQgazk8o:-MIsTMeN3GCysEW-bDlP:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/93ba28769428f894d934b4e1327125bc67c5adfc/scripts/dropDB.ts#L1)):*

Yes, let's remove `gen_test_cert.bat`.

Interesting, I assumed running `npm run version` wouldn't need `printVersion.js` to be executable because the file is being read by `node`, not executed directly, but I haven't tested on Linux yet.
</blockquote></details>

Done.

johnthagen (approved):
> Reviewed 2 of 6 files at r13.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/744)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #778, opened 2020-10-23
Cert server container

jmgrady:
Implement the ability for the certmgr container to be able to create SSL certificates as a proxy for remote/offline servers:
 - update documentation, both `certmgr/README.md` and docstrings
 - configure Nginx server to serve dummy pages for NUCs (so that certbot's http challenge will succeed)
 - add scripts to certmgr container:
    - aws.py for access to Amazon Web Services S3 bucket for certificates
    - cert_proxy_server.py to implement the CertProxyServer class.  CertProxyServer extends the LetsencryptCert class and adds creation/renewal of certficates as a proxy for offline devices.
 - remove deprecated bash shell versions of the certmgr scripts
 - move `func.py` to `utils.py`
 - add initial configuration files for the `nuc` group.  _These files will be used in a future PR._
 - merge in changes for backup and restore functions (affects some previous configuration items)
 - fixed generation of environment files for the backend, certmgr, and frontend containers - items that were integers instead of strings were not being included.
 - refactor `docker_install` role: make docker packages a list variable that can be overridden if needed.
 - update `scripts/docker_setup.py`

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)


imnasnainaec (commented):
> Reviewed 5 of 24 files at r5, 1 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: 6 of 21 files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[certmgr/README.md, line 33 at r6](https://reviewable.io/reviews/sillsdev/thecombine/778#-MKkBKI93Ec-0-6JJ9Jt:-MKkBKI93Ec-0-6JJ9Ju:b-hw1n2y) ([raw file](https://github.com/sillsdev/thecombine/blob/8c0d809e758fa9863055900f349beb9b478db436/certmgr/README.md#L33)):*
> ```Markdown
> | CERT_EMAIL        | Set to e-mail address for certificate expiration notices first                                                                                                       |
> | CERT_STAGING      | Primary domain for the certificate - used to specify location of certificate                                                                    |
> | CERT_ADDL_DOMAINS | A space separated list of domains additional for the certificate.                     |
> ```

How about "A space-separated list of additional domains for..."?
___
*[certmgr/README.md, line 35 at r6](https://reviewable.io/reviews/sillsdev/thecombine/778#-MKkD03B7hcKG-Z3-K4p:-MKkD03B7hcKG-Z3-K4q:brg80g9) ([raw file](https://github.com/sillsdev/thecombine/blob/8c0d809e758fa9863055900f349beb9b478db436/certmgr/README.md#L35)):*
> ```Markdown
> | CERT_ADDL_DOMAINS | A space separated list of domains additional for the certificate.                     |
> | SERVER_NAME       | Name of the server. Also used to specify the directory where the certificates are stored.                                                           |
> | MAX_CONNECT_TRIES | Number of times to check if the webserver is up before attempting to get a certificate from letsencrypt.                                        |
> ```

What about CERT_PROXY_RENEWAL and CERT_PROXY_DOMAINS?

johnthagen:
___
*[certmgr/README.md, line 104 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGsI8dBFbIXLOy5wgu:-MOGsI8eB3sshDVif-Sz:b-yxhd50) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/README.md#L104)):*
> ```Markdown
> proxy and uploads the certificates to an AWS S3 bucket
> 
> #### certclient.sh
> ```

Will this likely be a `.py` file? I'd suggest the name `cert_client.py`.

johnthagen:
___
*[certmgr/README.md, line 117 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGsVUT0VTLxGXvo1tT:-MOGsVUT0VTLxGXvo1tU:b-siz9i8) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/README.md#L117)):*
> ```Markdown
> or downloading objects from the bucket:
> 
> - `aws_s3_put(src, dest)`: Uploads the source file to the destination object. _The
> ```

In general, I think this kind of function documentation is easier to find and maintain as a docstring of the function itself. Otherwise it will be easy for this document to get out of sync with the source code.

johnthagen:
___
*[certmgr/README.md, line 131 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGsrZs9SfZm-VfkbHf:-MOGsrZt1o_nMo94lcvO:b-gnejz0) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/README.md#L131)):*
> ```Markdown
> #### utils.py
> 
> `utils.py` is a collection of utility functions:
> ```

Similar to the previous comment, I think a lot of this documentation for behavioral function could better live as module level or function level docstrings in the files themselves. That is where Python developers would typically look for them.

johnthagen:
___
*[certmgr/scripts/utils.py, line 17 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGtshS9fZxYfcQEVmj:-MOGtshS9fZxYfcQEVmk:b-8qykr9) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L17)):*
> ```Python
>     "SERVER_NAME": "",
>     "CERT_SELF_RENEWAL": 30,  # days before expriy
>     "PROXY_CERT_RENEWAL": 60,  # days before expiry
> ```

Typo in both `expriy` lines here -> `expire`

johnthagen:
___
*[certmgr/scripts/aws.py, line 21 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGuYFl0o6zkibk8GKr:-MOGuYFl0o6zkibk8GKs:bqidd1i) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L21)):*
> ```Python
>     aws_s3_uri = f"s3://{aws_s3_bucket}/{dest}"
>     print(f"AWS S3 put {src} to {dest}")
>     return os.system(f"aws s3 cp --profile {aws_s3_profile} {src} {aws_s3_uri}") == 0
> ```

In general, we should prefer `subprocess.run` over `os.system`. It's more flexible, powerful, and safe. You can pass the pieces of the command as a `List[str]` which allows it to be executed more safely (only the first item in the list is executed as a process, regardless of the other inputs, etc.)

https://docs.python.org/3/library/subprocess.html#subprocess.run

We could also potentially simplify our error handling by adding `check=True` to `subprocess.run`, since most of the time if one of these commands fails, we probably want to halt everything going on?

johnthagen:
___
*[certmgr/scripts/aws.py, line 34 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGvUsqARgwJMPK1Zqc:-MOGvUsqARgwJMPK1Zqd:b-ouluif) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L34)):*
> ```Python
>     aws_s3_bucket = get_setting("AWS_S3_CERT_LOC")
>     aws_s3_profile = get_setting("AWS_S3_PROFILE")
>     aws_s3_uri = f"s3://{aws_s3_bucket}/{src}"
> ```

These three lines look to be repeated in the function above. Could we define a private function that both of these functions call that encapsulates this work?

johnthagen:
___
*[certmgr/scripts/aws.py, line 42 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGvkWb8_8ERR8tNNH2:-MOGvkWb8_8ERR8tNNH3:b-qis7y3) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L42)):*
> ```Python
>     """Push all proxy certificates to AWS S3."""
>     cert_file_list = ("cert.pem", "chain.pem", "fullchain.pem", "privkey.pem")
>     domain_list: List[str] = get_setting("CERT_PROXY_DOMAINS").split()
> ```

Since `.split()` always returns a `List[str]` as specified in the std library, we can omit this `List[str]` type hint.

johnthagen:
___
*[certmgr/scripts/aws.py, line 44 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGy9ao7cXZ8INYoI_t:-MOGy9ao7cXZ8INYoI_u:b-5n5gvh) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L44)):*
> ```Python
>     domain_list: List[str] = get_setting("CERT_PROXY_DOMAINS").split()
>     for domain in domain_list:
>         cert_dir = Path(f"/etc/letsencrypt/live/{domain}")
> ```

The strings `"/etc/letsencrypt"`, `"/etc/letsencrypt/live"`, etc are duplicated several times across the following files. Could these paths be defined once as module level constants somewhere?

- cert_proxy_server.py
- aws.py
- letsencrypt_cert.py
- utils.py

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 46 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGzKO31jmTOOa1fYTu:-MOGzKO31jmTOOa1fYTv:bfknoj5) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L46)):*
> ```Python
>         """
>         super().create()
>         print("Called CertProxyServer.create()")
> ```

Do we still need this `print` function? In general, if I find myself leaning towards this kind of tracing (including class and function names in log message) I'd reach for the `logging` module which can add these kinds of things automatically: https://docs.python.org/3/library/logging.html#logrecord-attributes

https://stackoverflow.com/a/20112491

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 80 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGzoWO7lg9szLygVtB:-MOGzoWO7lg9szLygVtC:b-68cq7k) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L80)):*
> ```Python
>             aws_push_certs()
> 
>     def create_renew_hook(self) -> None:
> ```

Since this method does not use `self`, it can be converted into a `@staticmethod` to limit the scope of what it can mutate.

https://stackoverflow.com/a/735978

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 68 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH-SDiDTw5q-Q3Epyo:-MOH-SDiDTw5q-Q3Epyp:b-e559n9) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L68)):*
> ```Python
>            CERT_PROXY_DOMAINS variable.
>         """
>         domain_list: List[str] = get_setting("CERT_PROXY_DOMAINS").split()
> ```

Since `.split()` must return a `List[str]`, this type hint can be safely removed.

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 69 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH-YAEDC1UxbFHQZxt:-MOH-YAEDC1UxbFHQZxu:b-d2ek3) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L69)):*
> ```Python
>         """
>         domain_list: List[str] = get_setting("CERT_PROXY_DOMAINS").split()
>         cert_created: bool = False
> ```

Can remove `: bool` as that type is inferred by `False`

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 6 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH081S1vK7tE6AT1LE:-MOH081S1vK7tE6AT1LF:bqjt92x) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L6)):*
> ```
>  by fetching their certificates pushing them to an AWS S3 Bucket.
> ```

by fetching their certificates and pushing them

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 97 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH2UhP3YsKJKFwBoPZ:-MOH2UhP3YsKJKFwBoP_:b-b8cyt6) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L97)):*
> ```Python
>             renew_hook.symlink_to(hook_target)
>         else:
>             print(
> ```

We don't need to change this, but just as an FYI, the `logging` module has built in support for logging "warnings" differently: https://docs.python.org/3/library/logging.html#logging.warning. If we were to move to using `logging` we could use that.

But I don't think it's neccessary.

johnthagen:
___
*[certmgr/scripts/entrypoint.py, line 18 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH3did4C2GFYWRUUhz:-MOH3dieBXt3e6O50sk0:bczi2su) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L18)):*
> ```Python
> if __name__ == "__main__":
> 
>     mode_choices: Dict[str, Optional[BaseCert]] = {
> ```

I looks like we can remove this `Optional` as every entry has a concrete class instance associated with it.

johnthagen:
___
*[certmgr/scripts/entrypoint.py, line 17 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH4-Kp7Vu2jCWe1fJw:-MOH4-Kp7Vu2jCWe1fJx:b-w9jcgs) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L17)):*
> ```Python
> 
> if __name__ == "__main__":
> ```

Could remove this blank line.

johnthagen:
___
*[certmgr/scripts/entrypoint.py, line 40 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH4K3j1jcpvb3VO1_S:-MOH4K3j1jcpvb3VO1_T:be94w8i) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L40)):*
> ```Python
>     else:
>         print(f"Cannot run {cert_mode} mode")
>         sys.exit(99)
> ```

Should we return `1` here as the canonical "EXIT_FAILURE" status for POSIX?

https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 48 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH5OkT5upUxF_JaDtO:-MOH5OkUAaKDalwmS3gH:b-jddbwa) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L48)):*
> ```Python
>         this is successfull, then the symbolic link is moved to point to the
>         new certificate.
>         N O T E :
> ```

Other places in docstrings we type this as `NOTE:` without spaces for consistency.

johnthagen:
___
*[certmgr/scripts/self_signed_cert.py, line 11 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH5qH22YeHc784SgzO:-MOH5qH22YeHc784SgzP:b-b3fdmi) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/self_signed_cert.py#L11)):*
> ```Python
>     """Class for a Self-Signed SSL certificate."""
> 
>     def __init__(self, expire: int = 365, renew_before_expiry: int = 10) -> None:
> ```

We could define this as:

```
 def __init__(self, *, expire: int = 365, renew_before_expiry: int = 10) -> None:
```

To force callers to explicitly use keyword arguments for these parameters. The rationale, is that it's not super obvious what these parameters mean at the call site without the keywords.

```
a = SelfSignedCert(1, 0)
```

johnthagen:
___
*[certmgr/scripts/self_signed_cert.py, line 11 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH5qH22YeHc784SgzO:-MOH6Oo1-ty4bTmv9aD7:bhik8v7) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/self_signed_cert.py#L11)):*

We could define this as:

```
 def __init__(self, *, expire: int = 365, renew_before_expiry: int = 10) -> None:
```

To force callers to explicitly use keyword arguments for these parameters. The rationale, is that it's not super obvious what these parameters mean at the call site without the keywords.

```
a = SelfSignedCert(1, 0)
```
</blockquote></details>

See: https://www.python.org/dev/peps/pep-3102/

johnthagen:
___
*[certmgr/scripts/self_signed_cert.py, line 19 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH6hOD1WLyU8QYD44b:-MOH6hOD1WLyU8QYD44c:bk5temg) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/self_signed_cert.py#L19)):*
> ```Python
>         self.server_name = get_setting("SERVER_NAME")
>         self.expire = expire  # days
>         self.renew_before_expiry = renew_before_expiry  # days
> ```

Is "expiry" a term? I thought it was a typo, but now I see it used many places.

What does this word mean?

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 99 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH7BXF3RyuOqGovpR3:-MOH7BXF3RyuOqGovpR4:bdlquk4) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L99)):*
> ```Python
>         """
>         if domain_list is not None:
>             domain_args: str = "-d " + " -d ".join(domain_list)
> ```

Can be simplified to:

```
domain_args = f"-d {' -d '.join(domain_list)}"
```

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 108 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH7Lwk5vf94p_PNTih:-MOH7Lwk5vf94p_PNTii:b-ar8n7l) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L108)):*
> ```Python
>             staging_arg = "--staging" if self.staging else ""
> 
>             cert_cmd: str = (
> ```

Can remove this `: str` type hint as it is inferred from the RHS

johnthagen:
___
*[certmgr/scripts/utils.py, line 16 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH8INHBXIyjptLIbbU:-MOH8INHBXIyjptLIbbV:bi0rhfw) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L16)):*
> ```Python
>     "CERT_ADDL_DOMAINS": "",
>     "SERVER_NAME": "",
>     "CERT_SELF_RENEWAL": 30,  # days before expriy
> ```

expriy -> expiry

johnthagen:
___
*[certmgr/scripts/utils.py, line 8 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH8Add9HLkOYXJTkt0:-MOH8Add9HLkOYXJTkt1:blr7dhh) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L8)):*
> ```Python
> from typing import Dict, Optional, Union
> 
> env_defaults: Dict[str, Union[str, int, bool]] = {
> ```

I don't seen any `bool` variants so we can remove this part of the type hint.

johnthagen:
___
*[certmgr/scripts/utils.py, line 47 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH8mM57rzRZWnZOP2p:-MOH8mM57rzRZWnZOP2q:b-oz0ogk) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L47)):*
> ```Python
>     if dest.exists():
>         if dest.is_symlink():
>             link_target: str = os.readlink(dest)
> ```

We should be able to safely remove the `str` here.

johnthagen:
___
*[certmgr/scripts/aws.py, line 34 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGvUsqARgwJMPK1Zqc:-MOHB0aAExMSCpqIzI0c:bw0ko7i) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L34)):*

These three lines look to be repeated in the function above. Could we define a private function that both of these functions call that encapsulates this work?
</blockquote></details>

As noted, there are only actually 2 lines that can really be fully encapsulated (`aws_s3_profile` is used later in the function). A possible function that could be extracted stand-alone:

```
def _build_aws_s3_uri(path: str) -> str:
    aws_s3_bucket = get_setting("AWS_S3_CERT_LOC")
    return f"s3://{aws_s3_bucket}/{path}"
```

johnthagen:
___
*[docker_deploy/roles/combine_backup/templates/combine-backup.j2, line 112 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHDVl39JsgyeVj3gpo:-MOHDVl4BCA6chyTD8q2:bz1vmu2) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/docker_deploy/roles/combine_backup/templates/combine-backup.j2#L112)):*
> ```django
> #    need to specify full path because $PATH does not contain
> #    /usr/local/bin when run as a cron job
> /usr/local/bin/aws s3 cp ${BACKUP_FILE} ${AWS_FILE} --profile {{ aws_s3_profile }}
> ```

For ease of maintenance, we could consider porting these `bash` scripts in templates to Python. I think they could be a little easier to decompose into functions, have types, etc.

johnthagen:
___
*[docker_deploy/roles/combine_backup/templates/combine-restore.j2, line 107 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHDwry5civc8uXZeaa:-MOHDwry5civc8uXZeab:bt0ob0f) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/docker_deploy/roles/combine_backup/templates/combine-restore.j2#L107)):*
> ```django
> 
> if [ -z "${BACKUP}" ] ; then
>   aws_backup_list=( $(/usr/local/bin/aws s3 ls ${AWS_BACKUPS} --recursive --profile {{ aws_s3_profile }} | sed "s/.* backups\///") )
> ```

General question, is it possible that rather than Jinja-templating out these scripts, they could simply be constant but have more configuration values passed into them?

That could make them easier to review, but there likely are build time constraints that make this difficult?

johnthagen:
___
*[scripts/docker_setup.py, line 59 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHEcpa23friZpFepRV:-MOHEcpa23friZpFepRW:b-j6mp6m) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/scripts/docker_setup.py#L59)):*
> ```Python
>     its value.  If IMAGE_TAG is not defined, return "latest".
>     """
>     if "IMAGE_TAG" in os.environ:
> ```

Should we make this more project specific?

`"COMBINE_IMAGE_TAG"`?

johnthagen (changes requested):
> Reviewed 1 of 1 files at r4, 6 of 24 files at r5, 3 of 19 files at r7, 29 of 29 files at r9.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: all files reviewed, 26 unresolved discussions (waiting on @jmgrady)

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 46 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHH19nCVjlYs1N0PS8:-MOHH19nCVjlYs1N0PS9:bs4qiha) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L46)):*
> ```Python
>         This allows the Nginx webserver to start.  Once Nginx is up, then the
>         certificate can be requested using the webroot authentication method.  If
>         this is successfull, then the symbolic link is moved to point to the
> ```

Typo in "successfull"

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 31 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHHA7s0uiF-_n9V1Yr:-MOHHA7s0uiF-_n9V1Ys:b-maq8p3) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L31)):*
> ```Python
>         self.server_name = get_setting("SERVER_NAME")
>         self.email = get_setting("CERT_EMAIL")
>         self.max_connect_tries = int(get_setting("MAX_CONNECT_TRIES"))
> ```

We can change this to:

```
self.max_connect_tries: int = get_setting("MAX_CONNECT_TRIES")
```

In fact, each of these calls to `get_setting()` in this constructor should probably put a type hint on that is expected so that everywhere else that the attribute is used is properly type checked.

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 31 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHHdX34cj1el4jcRZA:-MOHHdX34cj1el4jcRZB:b8hfoeg) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L31)):*
> ```Python
>     """
> 
>     def __init__(self):
> ```

Should add explicit return type `-> None:`

johnthagen:
___
*[certmgr/scripts/utils.py, line 35 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHIC8q3x-hgrfAxYr7:-MOHIC8q3x-hgrfAxYr8:b5r2udw) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L35)):*
> ```Python
>     if env_var in env_defaults:
>         return env_defaults[env_var]
>     return None
> ```

We don't actually ever check for `None` in any of the usages of `get_setting`.

I would recommend to simplify the type structure to instead `raise MissingEnvironmentVariableError()` instead of returning `None`/`Optional` as that will fail quickly and then `None` won't bubble silently through our programs until it errors out somewhere later.

johnthagen (commented):
> Reviewed 2 of 2 files at r10.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: all files reviewed, 30 unresolved discussions (waiting on @jmgrady)

johnthagen:
A basic smoke test of the application running in Docker with self signed certificate worked properly.

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: all files reviewed, 28 unresolved discussions (waiting on @jmgrady and @johnthagen)

___
*[certmgr/README.md, line 104 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGsI8dBFbIXLOy5wgu:-MOH13mp6QaQgogWY63e:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/README.md#L104)):*

Will this likely be a `.py` file? I'd suggest the name `cert_client.py`.
</blockquote></details>

Done.
___
*[certmgr/README.md, line 117 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGsVUT0VTLxGXvo1tT:-MOH1PhbDl419Fn8MDQi:bld7fq5) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/README.md#L117)):*

In general, I think this kind of function documentation is easier to find and maintain as a docstring of the function itself. Otherwise it will be easy for this document to get out of sync with the source code.
</blockquote></details>

The details listed here are in the aws.py docstrings for the functions.  I will delete the detailed descriptions from README.md so that we don't have the maintenance issue that you raise.
___
*[certmgr/README.md, line 131 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGsrZs9SfZm-VfkbHf:-MOH2-t-AOgp0d-esGgj:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/README.md#L131)):*

Similar to the previous comment, I think a lot of this documentation for behavioral function could better live as module level or function level docstrings in the files themselves. That is where Python developers would typically look for them.
</blockquote></details>

Done.
___
*[certmgr/scripts/aws.py, line 34 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGvUsqARgwJMPK1Zqc:-MOHCufE-DxGrF2nxFS_:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L34)):*

As noted, there are only actually 2 lines that can really be fully encapsulated (`aws_s3_profile` is used later in the function). A possible function that could be extracted stand-alone:

```
def _build_aws_s3_uri(path: str) -> str:
    aws_s3_bucket = get_setting("AWS_S3_CERT_LOC")
    return f"s3://{aws_s3_bucket}/{path}"
```
</blockquote></details>

Done.
___
*[certmgr/scripts/aws.py, line 42 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGvkWb8_8ERR8tNNH2:-MOHD6JiCiP25Uw6WBBs:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L42)):*

Since `.split()` always returns a `List[str]` as specified in the std library, we can omit this `List[str]` type hint.
</blockquote></details>

Done.
___
*[certmgr/scripts/aws.py, line 44 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGy9ao7cXZ8INYoI_t:-MOHLQJL-DDDjrJf4ypa:b5sa64k) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L44)):*

The strings `"/etc/letsencrypt"`, `"/etc/letsencrypt/live"`, etc are duplicated several times across the following files. Could these paths be defined once as module level constants somewhere?

- cert_proxy_server.py
- aws.py
- letsencrypt_cert.py
- utils.py
</blockquote></details>

I have been able to reduce the occurrences significantly by making it an attribute of LetsEncryptCert.  Is there a recommended pattern for creating module level constants?  I found this reference: https://www.oreilly.com/library/view/python-cookbook/0596001673/ch05s16.html
Is this what you had in mind?
___
*[certmgr/scripts/cert_proxy_server.py, line 6 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH081S1vK7tE6AT1LE:-MOHM4aq-xUebugwM5PH:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L6)):*

> ```
>  by fetching their certificates pushing them to an AWS S3 Bucket.
> ```

by fetching their certificates and pushing them
</blockquote></details>

Done.
___
*[certmgr/scripts/cert_proxy_server.py, line 31 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHHdX34cj1el4jcRZA:-MOHM57M-ZIyqNxw7SSk:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L31)):*

Should add explicit return type `-> None:`
</blockquote></details>

Done.
___
*[certmgr/scripts/cert_proxy_server.py, line 46 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGzKO31jmTOOa1fYTu:-MOHMMPC2qm5_3zj1xJs:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L46)):*

Do we still need this `print` function? In general, if I find myself leaning towards this kind of tracing (including class and function names in log message) I'd reach for the `logging` module which can add these kinds of things automatically: https://docs.python.org/3/library/logging.html#logrecord-attributes

https://stackoverflow.com/a/20112491
</blockquote></details>

Done.
___
*[certmgr/scripts/cert_proxy_server.py, line 68 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH-SDiDTw5q-Q3Epyo:-MOHMq0t6MtU1dpIrlRI:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L68)):*

Since `.split()` must return a `List[str]`, this type hint can be safely removed.
</blockquote></details>

Done.
___
*[certmgr/scripts/cert_proxy_server.py, line 69 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH-YAEDC1UxbFHQZxt:-MOHMuJA8NMEMpTOnJsA:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L69)):*

Can remove `: bool` as that type is inferred by `False`
</blockquote></details>

Done.
___
*[certmgr/scripts/cert_proxy_server.py, line 80 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGzoWO7lg9szLygVtB:-MOHN0k_D03iI0l3d6tE:b-elxp8d) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L80)):*

Since this method does not use `self`, it can be converted into a `@staticmethod` to limit the scope of what it can mutate.

https://stackoverflow.com/a/735978
</blockquote></details>

It uses `self` now! (Well, `super()` actually).  In order to address all the instances of `"/etc/letsencrypt"` I made it an attribute of the LetsEncryptCert class so now this method looks up the directory in the parent class.  Of course, if your method for creating module constants is better, then I will make this a static function.
___
*[certmgr/scripts/entrypoint.py, line 17 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH4-Kp7Vu2jCWe1fJw:-MOHO7LY2s_BPU2AsNSo:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L17)):*

Could remove this blank line.
</blockquote></details>

Done.
___
*[certmgr/scripts/entrypoint.py, line 18 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH3did4C2GFYWRUUhz:-MOHOCFzBxHz63oAqbOG:bh0ssym) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L18)):*

I looks like we can remove this `Optional` as every entry has a concrete class instance associated with it.
</blockquote></details>

I thought I would need Optional so that this will work:
```
    ...
    cert_obj = mode_choices.get(cert_mode, None)

    if cert_obj is not None:
```
in case someone tried to call something like `mode_choices["foo"]`
___
*[certmgr/scripts/letsencrypt_cert.py, line 31 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHHA7s0uiF-_n9V1Yr:-MOHPXSYAWLReAEGHwgj:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L31)):*

We can change this to:

```
self.max_connect_tries: int = get_setting("MAX_CONNECT_TRIES")
```

In fact, each of these calls to `get_setting()` in this constructor should probably put a type hint on that is expected so that everywhere else that the attribute is used is properly type checked.
</blockquote></details>

Done.
___
*[certmgr/scripts/letsencrypt_cert.py, line 46 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHH19nCVjlYs1N0PS8:-MOHPjKc6ZDd7MNsDaXu:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L46)):*

Typo in "successfull"
</blockquote></details>

Done.
___
*[certmgr/scripts/letsencrypt_cert.py, line 48 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH5OkT5upUxF_JaDtO:-MOHPopBCx8fUJC8mnG3:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L48)):*

Other places in docstrings we type this as `NOTE:` without spaces for consistency.
</blockquote></details>

Done.
___
*[certmgr/scripts/letsencrypt_cert.py, line 99 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH7BXF3RyuOqGovpR3:-MOHQFPl3DKnOn7j41Vu:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L99)):*

Can be simplified to:

```
domain_args = f"-d {' -d '.join(domain_list)}"
```
</blockquote></details>

Done.
___
*[certmgr/scripts/letsencrypt_cert.py, line 108 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH7Lwk5vf94p_PNTih:-MOHQLVgBT15272V0UDP:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/letsencrypt_cert.py#L108)):*

Can remove this `: str` type hint as it is inferred from the RHS
</blockquote></details>

Done.
___
*[certmgr/scripts/self_signed_cert.py, line 11 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH5qH22YeHc784SgzO:-MOHR8gd6bN7CU3qTEOF:bi949pr) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/self_signed_cert.py#L11)):*

See: https://www.python.org/dev/peps/pep-3102/
</blockquote></details>

Much better! Done.
___
*[certmgr/scripts/self_signed_cert.py, line 19 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH6hOD1WLyU8QYD44b:-MOHRDE622rJ8TmgtpFK:b-783e3v) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/self_signed_cert.py#L19)):*

Is "expiry" a term? I thought it was a typo, but now I see it used many places.

What does this word mean?
</blockquote></details>

It is the noun form of the verb 'expire'
___
*[docker_deploy/roles/combine_backup/templates/combine-backup.j2, line 112 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHDVl39JsgyeVj3gpo:-MOHV5rd9ugeYMaD5AHG:b47ic6) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/docker_deploy/roles/combine_backup/templates/combine-backup.j2#L112)):*

For ease of maintenance, we could consider porting these `bash` scripts in templates to Python. I think they could be a little easier to decompose into functions, have types, etc.
</blockquote></details>

I disagree here.  I think that the script would be primarily calls to os.system or subprocess.run so I think that `bash` is the right tool for this problem.  (I'm willing to be convinced otherwise, though).
___
*[docker_deploy/roles/combine_backup/templates/combine-restore.j2, line 107 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHDwry5civc8uXZeaa:-MOHViE30AxBeoo297dA:bysclqc) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/docker_deploy/roles/combine_backup/templates/combine-restore.j2#L107)):*

General question, is it possible that rather than Jinja-templating out these scripts, they could simply be constant but have more configuration values passed into them?

That could make them easier to review, but there likely are build time constraints that make this difficult?
</blockquote></details>

I am using Jinja to fill in some of the machine specific details or current Combine configuration values.  I can look into specifying these values as environment variables (esp. for a cron job) or as command line options - I've created Issue #860 for this.

One thing that makes reading the script a _lot_ easier is that Atom has a package called `shebang-set-grammar` which uses the first line of the script to determine the programming language instead of the filename extension.  I have looked for something like this for VS Code but most of the search results are about building my own extension. :-(
___
*[scripts/docker_setup.py, line 59 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHEcpa23friZpFepRV:-MOHd6J99vBesCgSJglU:b-41g22r) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/scripts/docker_setup.py#L59)):*

Should we make this more project specific?

`"COMBINE_IMAGE_TAG"`?
</blockquote></details>

I don't know that it is necessary.  The IMAGE_TAG is never set as a general environment variable for a user on the system - it is stored in the .env file with the docker-compose.yml file.  I can create an issue for it if you like.  It would not be hard to do but would impact the CI/CD as well as our code.  Maybe that's more incentive to move to GitActions! ;-)
___
*[certmgr/scripts/utils.py, line 8 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH8Add9HLkOYXJTkt0:-MOHRu2VCMlOhWb0w-z5:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L8)):*

I don't seen any `bool` variants so we can remove this part of the type hint.
</blockquote></details>

Done.
___
*[certmgr/scripts/utils.py, line 16 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH8INHBXIyjptLIbbU:-MOHS0RaC9qWuNtEODAK:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L16)):*

expriy -> expiry
</blockquote></details>

Done.
___
*[certmgr/scripts/utils.py, line 17 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGtshS9fZxYfcQEVmj:-MOHS1v_8qf6g-UDZ3bJ:b8ystn1) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L17)):*

Typo in both `expriy` lines here -> `expire`
</blockquote></details>

See above.  "expiry" is correct.  In addition, it is specifically for setting the configuration value of `renew_before_expiry` in `/etc/letsencrypt/renewal/<domainname>.conf` (if that helps any).
___
*[certmgr/scripts/utils.py, line 35 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHIC8q3x-hgrfAxYr7:-MOHUg5KC_SvS1seJQR9:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L35)):*

We don't actually ever check for `None` in any of the usages of `get_setting`.

I would recommend to simplify the type structure to instead `raise MissingEnvironmentVariableError()` instead of returning `None`/`Optional` as that will fail quickly and then `None` won't bubble silently through our programs until it errors out somewhere later.
</blockquote></details>

Done.
___
*[certmgr/scripts/utils.py, line 47 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH8mM57rzRZWnZOP2p:-MOHUqeT23FxjybVfvL3:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/utils.py#L47)):*

We should be able to safely remove the `str` here.
</blockquote></details>

Done.

johnthagen:
___
*[certmgr/scripts/aws.py, line 44 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGy9ao7cXZ8INYoI_t:-MOHgyqZ5O_CPgouN8SG:b7ef0gq) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L44)):*

I have been able to reduce the occurrences significantly by making it an attribute of LetsEncryptCert.  Is there a recommended pattern for creating module level constants?  I found this reference: https://www.oreilly.com/library/view/python-cookbook/0596001673/ch05s16.html
Is this what you had in mind?
</blockquote></details>

The convention if it's a constant value would be something like at the top of a module:

```
LETSENCRYPT_CONFIG_DIR = Path("/etc/letsencrypt")
```

And then other modules would import this constant:

```
from letsencrypt_cert import LETSENCRYPT_CONFIG_DIR
```

It could also be declared as a constant associated with a class and used like:

```
LetsEncryptCert.LETSENCRYPT_CONFIG_DIR
```

just depending on where you think the information would best live.

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 32 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHhtWWCRfTdxq_SMGG:-MOHhtWWCRfTdxq_SMGH:bb5v4m) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L32)):*
> ```Python
>         self.max_connect_tries: int = get_setting("MAX_CONNECT_TRIES")
>         self.staging = get_setting("CERT_STAGING") != "0"
>         self.le_dir = Path("/etc/letsencrypt")
> ```

Since this is always a constant value, it would be best to make it a "static" class attribute rather than an instance attribute. That way, you don't need a reference to `self` to access it (and it could even be accessed outside the class as well).

You can do this by moving it outside the constructor:

```
class LetsEncryptCert(BaseCert):
    """SSL Certificate class to create and renew certs from Let's Encrypt."""

    LE_DIR = Path("/etc/letsencrypt")

    def __init__ ...
```

Python allows you to access static class attributes with `self` for convenience if you are in a method, but this will allow this to be accessed from a `staticmethod` or without an instance to this class. This allows usages of it without signaling the need for (potentially) mutable access to the instance. (Since Python doesn't have the equivalent of a `const` method from C++ or taking `self ` without `mut` as in Rust).

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 80 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGzoWO7lg9szLygVtB:-MOHidkj1pBJVy5Yl_Dg:b-hs65mp) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L80)):*

It uses `self` now! (Well, `super()` actually).  In order to address all the instances of `"/etc/letsencrypt"` I made it an attribute of the LetsEncryptCert class so now this method looks up the directory in the parent class.  Of course, if your method for creating module constants is better, then I will make this a static function.
</blockquote></details>

I commented below, but yes ideally we should try to make things static when possible to limit their scope. Making a class attribute (as noted below) or a module attribute would be better for constants. It also uses a little bit less memory because each instance does't need to allocate a spot for the same constant.

johnthagen:
___
*[certmgr/scripts/entrypoint.py, line 18 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH3did4C2GFYWRUUhz:-MOHj5zz7IC4S5cQ0-Q2:bgs0rzc) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L18)):*

I thought I would need Optional so that this will work:
```
    ...
    cert_obj = mode_choices.get(cert_mode, None)

    if cert_obj is not None:
```
in case someone tried to call something like `mode_choices["foo"]`
</blockquote></details>

The `.get()` method on `dict`s is what can return `None` (or whatever default value you'd like.

So `cert_obj` in the example above would be `Optional` (since we are using `None` as the default), but `mode_choices` itself does not have `Optional` interior values.

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 90 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjbJ07DzK9Houc2m7:-MOHjbJ07DzK9Houc2m8:b-qke3li) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L90)):*
> ```Python
>         os.system("certbot renew")
> 
>     def get_le_dir(self) -> Path:
> ```

This can be removed if we make a constant static class attribute for `LE_DIR`.

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 152 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjjFZ-_7LOJCUbuwT:-MOHjjF_9JHggbh1q4TS:b-8b0dzn) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L152)):*
> ```Python
>         return False
> 
>     def update_renew_before_expiry(self, domain: str, renew_before_expiry_period: int) -> None:
> ```

This can be made `staticmethod` once we make a static `LE_DIR`.

johnthagen:
___
*[docker_deploy/roles/combine_backup/templates/combine-restore.j2, line 107 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHDwry5civc8uXZeaa:-MOHkNd82nb4TEieyLU0:blteee8) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/docker_deploy/roles/combine_backup/templates/combine-restore.j2#L107)):*

I am using Jinja to fill in some of the machine specific details or current Combine configuration values.  I can look into specifying these values as environment variables (esp. for a cron job) or as command line options - I've created Issue #860 for this.

One thing that makes reading the script a _lot_ easier is that Atom has a package called `shebang-set-grammar` which uses the first line of the script to determine the programming language instead of the filename extension.  I have looked for something like this for VS Code but most of the search results are about building my own extension. :-(
</blockquote></details>

Yeah, Rider syntax highlights them fine actually. It's more trying to understand the "inputs" and control flow that was a bit more difficult for me at first.

johnthagen (changes requested):
> Reviewed 6 of 8 files at r11.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: 34 of 36 files reviewed, 7 unresolved discussions (waiting on @jmgrady and @johnthagen)

johnthagen:
___
*[certmgr/scripts/cert_proxy_server.py, line 84 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHiwRQ-lAleb3dbnqE:-MOIe4pX8JdcJIqENRuh:bb6ywmx) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/cert_proxy_server.py#L84)):*

Replace with constant lookup (either static class attribute or module constant).
</blockquote></details>

LGTM

johnthagen:
___
*[certmgr/scripts/entrypoint.py, line 18 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH3did4C2GFYWRUUhz:-MOIe92OBhZxk626kt4F:bb6ywmx) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L18)):*

The `.get()` method on `dict`s is what can return `None` (or whatever default value you'd like.

So `cert_obj` in the example above would be `Optional` (since we are using `None` as the default), but `mode_choices` itself does not have `Optional` interior values.
</blockquote></details>

LGTM

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 32 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHhtWWCRfTdxq_SMGG:-MOIeCFX4rJ2LoVcNe7k:bb6ywmx) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L32)):*

Since this is always a constant value, it would be best to make it a "static" class attribute rather than an instance attribute. That way, you don't need a reference to `self` to access it (and it could even be accessed outside the class as well).

You can do this by moving it outside the constructor:

```
class LetsEncryptCert(BaseCert):
    """SSL Certificate class to create and renew certs from Let's Encrypt."""

    LE_DIR = Path("/etc/letsencrypt")

    def __init__ ...
```

Python allows you to access static class attributes with `self` for convenience if you are in a method, but this will allow this to be accessed from a `staticmethod` or without an instance to this class. This allows usages of it without signaling the need for (potentially) mutable access to the instance. (Since Python doesn't have the equivalent of a `const` method from C++ or taking `self ` without `mut` as in Rust).
</blockquote></details>

LGTM

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 90 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjbJ07DzK9Houc2m7:-MOIeZPg4tqvB9L-k_zU:bb6ywmx) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L90)):*

This can be removed if we make a constant static class attribute for `LE_DIR`.
</blockquote></details>

LGTM

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 152 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjjFZ-_7LOJCUbuwT:-MOIeafa3V7zV3Se9FZW:bb6ywmx) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L152)):*

This can be made `staticmethod` once we make a static `LE_DIR`.
</blockquote></details>

LGTM

johnthagen (approved):
> Reviewed 1 of 8 files at r11, 1 of 5 files at r12, 6 of 8 files at r13, 2 of 2 files at r14.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: 36 of 37 files reviewed, all discussions resolved (waiting on @johnthagen)

jmgrady (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778#-:-MOIsbKF-QNmtXXqjIWW:bnfp4nl)* status: 36 of 37 files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[certmgr/scripts/aws.py, line 44 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGy9ao7cXZ8INYoI_t:-MOIVYvK6yLaIbHZW4NI:bidijdp) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/aws.py#L44)):*

The convention if it's a constant value would be something like at the top of a module:

```
LETSENCRYPT_CONFIG_DIR = Path("/etc/letsencrypt")
```

And then other modules would import this constant:

```
from letsencrypt_cert import LETSENCRYPT_CONFIG_DIR
```

It could also be declared as a constant associated with a class and used like:

```
LetsEncryptCert.LETSENCRYPT_CONFIG_DIR
```

just depending on where you think the information would best live.
</blockquote></details>

Done.
I made it part of LetsEncryptCert.
___
*[certmgr/scripts/cert_proxy_server.py, line 80 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOGzoWO7lg9szLygVtB:-MOIWtpF3Ki-QhVcsr_h:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/cert_proxy_server.py#L80)):*

I commented below, but yes ideally we should try to make things static when possible to limit their scope. Making a class attribute (as noted below) or a module attribute would be better for constants. It also uses a little bit less memory because each instance does't need to allocate a spot for the same constant.
</blockquote></details>

Done.
___
*[certmgr/scripts/cert_proxy_server.py, line 84 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHiwRQ-lAleb3dbnqE:-MOIXARp6wXIAo5vAdC1:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/cert_proxy_server.py#L84)):*

Replace with constant lookup (either static class attribute or module constant).
</blockquote></details>

Done.
___
*[certmgr/scripts/entrypoint.py, line 18 at r9](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOH3did4C2GFYWRUUhz:-MOICMPq2QGAl4M-9t9q:b-1fe1p4) ([raw file](https://github.com/sillsdev/thecombine/blob/cbf0b7625b291f3fb3f178655fc02db6a07dc67e/certmgr/scripts/entrypoint.py#L18)):*

The `.get()` method on `dict`s is what can return `None` (or whatever default value you'd like.

So `cert_obj` in the example above would be `Optional` (since we are using `None` as the default), but `mode_choices` itself does not have `Optional` interior values.
</blockquote></details>

That makes sense!  Thank.
Done.
___
*[certmgr/scripts/letsencrypt_cert.py, line 32 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHhtWWCRfTdxq_SMGG:-MOIXEubEEirTU-UsOCL:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L32)):*

Since this is always a constant value, it would be best to make it a "static" class attribute rather than an instance attribute. That way, you don't need a reference to `self` to access it (and it could even be accessed outside the class as well).

You can do this by moving it outside the constructor:

```
class LetsEncryptCert(BaseCert):
    """SSL Certificate class to create and renew certs from Let's Encrypt."""

    LE_DIR = Path("/etc/letsencrypt")

    def __init__ ...
```

Python allows you to access static class attributes with `self` for convenience if you are in a method, but this will allow this to be accessed from a `staticmethod` or without an instance to this class. This allows usages of it without signaling the need for (potentially) mutable access to the instance. (Since Python doesn't have the equivalent of a `const` method from C++ or taking `self ` without `mut` as in Rust).
</blockquote></details>

Done.
___
*[certmgr/scripts/letsencrypt_cert.py, line 90 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjbJ07DzK9Houc2m7:-MOID5BWCLW1AQ1zO8ep:b-rhbque) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L90)):*

This can be removed if we make a constant static class attribute for `LE_DIR`.
</blockquote></details>

I don't think that it can.  when `certbot` is run with the `renew` command, it checks all of the Let's Encrypt certificates that were issued to this server.  If any of them are expiring soon (i.e. if there are less than `renew_before_expiry` days left before the certificates expire) then `certbot` will request a new certificate from Let's Encrypt.  LE_DIR is not used.
___
*[certmgr/scripts/letsencrypt_cert.py, line 152 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjjFZ-_7LOJCUbuwT:-MOINyT-8mKG5uNbscbQ:b-tsubje) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L152)):*

This can be made `staticmethod` once we make a static `LE_DIR`.
</blockquote></details>

It still references self.email and self.staging.

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 90 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjbJ07DzK9Houc2m7:-MON3DdN9-xaJmkBY7dX:b-cr42y0) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L90)):*

I don't think that it can.  when `certbot` is run with the `renew` command, it checks all of the Let's Encrypt certificates that were issued to this server.  If any of them are expiring soon (i.e. if there are less than `renew_before_expiry` days left before the certificates expire) then `certbot` will request a new certificate from Let's Encrypt.  LE_DIR is not used.
</blockquote></details>

(Yeah I think this comment used to be for a now-deleted line of code. I didn't mean to suggest removing this `os.system` call.

johnthagen:
___
*[certmgr/scripts/letsencrypt_cert.py, line 152 at r11](https://reviewable.io/reviews/sillsdev/thecombine/778#-MOHjjFZ-_7LOJCUbuwT:-MON3NME12r9eHt5mMGn:b-tctuc4) ([raw file](https://github.com/sillsdev/thecombine/blob/d15abbed2b6494858bea68fcc3455076d4fd39db/certmgr/scripts/letsencrypt_cert.py#L152)):*

It still references self.email and self.staging.
</blockquote></details>

Similar comment. You addressed the original comment, but now that code has been removed, the comment has been solved.

johnthagen (approved):
> Reviewed 1 of 1 files at r15.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/778)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #811, opened 2020-11-13
Reduce peak memory usage during Lift Export


johnthagen:
Currently when the backend exports a Lift file, it:

- Generates a zip file temporarily on disk (can be many 100s of MB)
- It reads this into memory
- It base64 decodes the contents in memory (which expands it ~30%)
- It sends this to the client in an HTTP response

This leads to peak memory usage of about (export_size x 2.3). It has been observed that for a 600MB export, system memory utilization increased to 2.2GB.

One option is to use NGINX `sendfile` to allow NGINX to efficiently stream the file back to the client and not load it into the backend memory:

- https://nginx.org/en/docs/http/ngx_http_core_module.html#sendfile

johnthagen:
Related to #688 / #694

PR #814, opened 2020-11-13
Avoid Base64 encoding lift export files

johnthagen:
Helps address #811

- This should remove about 60% of the peak memory usage during Lift export.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/814)


johnthagen:
The frontend portion of this still needs to be completed.

Currently fails here:


```
ExportProjectActions.tsx:35 GET data:application/zip;base64,PK%03%04%14%00%00%00%08%00%EF%BF%BDStQ%EF%BF%BD net::ERR_INVALID_URL
```

```ts
export function asyncDownloadExport(projectId?: string) {
  return async (
    dispatch: ThunkDispatch<StoreState, any, ExportProjectAction>
  ) => {
    return downloadLift(projectId)
      .then((fileString) => {
        return fetch(fileString)
          .then((file) => {
            dispatch(success(projectId));
            return file.blob();
          })
          .catch(() => {
            dispatch(failure(projectId));
          });
      })
      .catch(() => {
        dispatch(failure(projectId));
      });
  };
}
```

imnasnainaec (approved):
> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/814)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/backend/index.tsx, line 365 at r9](https://reviewable.io/reviews/sillsdev/thecombine/814#-MMb205tAeC6HE_dJLy6:-MMb205tAeC6HE_dJLy7:bdlp19r) ([raw file](https://github.com/sillsdev/thecombine/blob/d5feb9fecb74797a1ac85b82fcae0cd01cc4baa5/src/backend/index.tsx#L365)):*
> ```
>   return window.URL.createObjectURL(
>     new Blob([resp.request.response], { type: "application/zip" })
>   );
> ```
I think this is redundant with some of what is happening in `ExportProjectActions` and `ExportProjectButton`, but that will be easier to simplify within #817 after we merge this in.

PR #819, opened 2020-11-18
Split lift export function in backend

imnasnainaec:
Follow-up to #812 
Prerequisite to #817

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/819)

PR #820, opened 2020-11-20
Fix memory leak in Lift Export

johnthagen:
Follow on to #819 

Relate to #814 

- Fix memory leak in Lift export where Lift exported files are never removed from an internal dictionary
- Reduce medium-term memory usage of lift export by 33% by only storing binary contents of export rather than base64 encoded contents before download

Note: There is still a possibility of a memory leak if a user exports a project but never downloads it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/820)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/820)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #823, opened 2020-11-20
Only cache the path to exported lift files rather than contents

### LINKED ISSUE #822: Memory leak in Lift Export if user never downloads export
If a user exports a Lift project but does not Download it, the entire contents of the Lift export (which could be 100's MB) will be stored in memory until the user Downloads it. If it is never downloaded, the memory is leaked permanently.

johnthagen:
Closes #822 

Worst case, if the user never downloads the project, only the file path to the export is leaked, rather than the entire contents of the export.


imnasnainaec (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/823)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #830, opened 2020-11-30
Set ecr profile read only

jmgrady:
Set the profile used to access AWS Elastic Container Registry (ECR) to `ecr_read_only`.

Servers only need to download docker images; they do not need to upload new ones.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/830)


jmgrady:
Merge from master first.

PR #838, opened 2020-12-02
Add wrapper inside tooltip around functional component.

imnasnainaec:
Workaround that will need to be generalized if we use Tooltip around functional components in more places.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/838)


imnasnainaec:
Replaced by #839.

PR #839, opened 2020-12-02
Move Tooltip inside RecorderIcon.

imnasnainaec:
Replaces #838.

The `ref` property is not implicit with a functional component as it is for a traditional component.
Thus, `Tooltip` cannot be used around a functional component, because it tries to pass its `ref` to its children.
Previously in `AudioRecorder`, we had `Tooltip` wrapped around the functional component `RecorderIcon`.
To fix the associated warnings, this pr moves `Tooltip` to within `RecorderIcon`.
Should we ever need tooltip around a functional component: https://material-ui.com/guides/composition/#caveat-with-refs

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/839)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/839)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #902, opened 2021-01-07
Cleanup loadGoalData and blacklist building.

imnasnainaec:
`loadGoalData()` blocks any use of words that are already in a suggested duplicate set, as it should. It also blocks words that were seen in a singleton set or a blacklist set, preventing them from being compared to other words. This pr removes those two restrictions.

The latter restriction was because `getNextDups()` has some sets that are strict subsets of previous sets. If you blacklisted a set, you shouldn't have to consider its subsets. To target that problem specifically, this pr updates the blacklisting function to blacklist not just a specified set but also its subsets.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/902)

### DIFF
--- a/src/components/GoalTimeline/GoalsActions.tsx
+++ b/src/components/GoalTimeline/GoalsActions.tsx
@@ -8,7 +8,6 @@ import { CreateStrWordInv } from "../../goals/CreateStrWordInv/CreateStrWordInv"
 import { HandleFlags } from "../../goals/HandleFlags/HandleFlags";
 import DupFinder from "../../goals/MergeDupGoal/DuplicateFinder/DuplicateFinder";
 import { MergeDupData, MergeDups } from "../../goals/MergeDupGoal/MergeDups";
-import { Hash } from "../../goals/MergeDupGoal/MergeDupStep/MergeDupsTree";
 import {
   MergeTreeAction,
   refreshWords,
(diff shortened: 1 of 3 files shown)


jasonleenaylor (approved):
It looks good. I don't think it will generate enough content to run into problems in local storage. I do wonder though if we ought to be offloading some of this into the backend. Perhaps by storing it in the goal info. We do have an eventual goal of displaying some information about what the user accomplished. e.g. X words merged, Y senses merged, Z senses promoted to entry. That enhancement can come later.

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/902#-:-MQTnohA0F7vMnX_4VYr:bc6qbtq)* status: :shipit: complete! all files reviewed, all discussions resolved

jasonleenaylor (commented):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/902)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

jasonleenaylor (commented):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/902#-:-MQXc5jhBHLNONcqdzDO:b-uwkowd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

Issue #927, opened 2021-01-21
LIFT import/export fails for dev on mac


imnasnainaec:
Recreate issue:
* Follow https://github.com/sillsdev/TheCombine#getting-started-with-development on macOS 10.15
* From within the project folder, run `npm run test-backend -- --filter FullyQualifiedName~Backend.Tests.Controllers.LiftControllerTests`

```
Test run for ~/TheCombine/Backend.Tests/bin/Debug/netcoreapp3.1/Backend.Tests.dll(.NETCoreApp,Version=v3.1)
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.
  X TestExportDeleted [105ms]
  Error Message:
   SIL.PlatformUtilities.NativeException : An error with the number, 13, ocurred.
  Stack Trace:
     at SIL.Threading.GlobalMutex.LinuxGlobalMutexAdapter.Init(Boolean initiallyOwned)
   at SIL.Threading.GlobalMutex.Initialize()
   at SIL.WritingSystems.Sldr.Initialize(Boolean offlineMode, String sldrCachePath, DateTime embeddedAllTagsTime)
   at SIL.WritingSystems.Sldr.Initialize(Boolean offlineMode, String sldrCachePath)
   at SIL.WritingSystems.Sldr.Initialize(Boolean offlineMode)
   at BackendFramework.Services.LiftService..ctor() in ~/TheCombine/Backend/Services/LiftApiServices.cs:line 98
   at Backend.Tests.Controllers.LiftControllerTests.Setup() in ~/TheCombine/Backend.Tests/Controllers/LiftControllerTests.cs:line 39
  X TestRoundtrip(Backend.Tests.Controllers.LiftControllerTests+RoundTripObj) [2s 48ms]
  Error Message:
     Expected: True
  But was:  False

  Stack Trace:
     at Backend.Tests.Controllers.LiftControllerTests.TestRoundtrip(RoundTripObj roundTripObj) in ~/TheCombine/Backend.Tests/Controllers/LiftControllerTests.cs:line 315
```

The SIL GlobalMutex has been written for Windows and Linux, and defaults to Linux when Windows isn't detected:
[...]

PR #934, opened 2021-01-22
Don't depend LiftImported=true on LdmlImport.

imnasnainaec:
There are presently issues with the LdmlImport on mac (#927), which put me in a strange situation. The words would successfully import into a project, but the final step of the import would fail, so `project.LiftImported` was never marked as true.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/934)


johnthagen:
___
*[Backend/Controllers/LiftController.cs, line 168 at r2](https://reviewable.io/reviews/sillsdev/thecombine/934#-MRf1Hv247FsXEu8zb1W:-MRf1Hv247FsXEu8zb1X:bmawasx) ([raw file](https://github.com/sillsdev/thecombine/blob/f85ed9bb694a9b784df7cc33fd6633fd27f50399/Backend/Controllers/LiftController.cs#L168)):*
> ```Smalltalk
>                     return new NotFoundObjectResult(projectId);
>                 }
>                 project.LiftImported = true;
> ```

It seems like with this change, if Lift import fails, we still mark the `Project` as having imported, so the user would never be able to try to add a fixed Lift file again?

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/934)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @johnthagen)

___
*[Backend/Controllers/LiftController.cs, line 168 at r2](https://reviewable.io/reviews/sillsdev/thecombine/934#-MRf1Hv247FsXEu8zb1W:-MRf8XjP31YyqpUozZLM:b-dx9ef9) ([raw file](https://github.com/sillsdev/thecombine/blob/f85ed9bb694a9b784df7cc33fd6633fd27f50399/Backend/Controllers/LiftController.cs#L168)):*

It seems like with this change, if Lift import fails, we still mark the `Project` as having imported, so the user would never be able to try to add a fixed Lift file again?
</blockquote></details>

It only marks as imported after ``parser.ReadLiftFile()`` has imported the words into the project without error.

johnthagen:
___
*[Backend/Controllers/LiftController.cs, line 168 at r2](https://reviewable.io/reviews/sillsdev/thecombine/934#-MRf1Hv247FsXEu8zb1W:-MRjS1xiA4Awmms5wJFa:b5q93sp) ([raw file](https://github.com/sillsdev/thecombine/blob/6e1588204ed54beee909b885ea46dcbc68083f61/Backend/Controllers/LiftController.cs#L168)):*

It only marks as imported after ``parser.ReadLiftFile()`` has imported the words into the project without error.
</blockquote></details>

What if `_liftService.LdmlImport()` throws though? To me it seems like we should only set `LiftImported` to `true` after everything has fully completed successfully. Or is changing this behavior part of what you are needing to fix for macOS?

johnthagen (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/934)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/934)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec and @johnthagen)

___
*[Backend/Controllers/LiftController.cs, line 168 at r2](https://reviewable.io/reviews/sillsdev/thecombine/934#-MRf1Hv247FsXEu8zb1W:-MRjplruDIR0lZ3N8EAq:b6l69zn) ([raw file](https://github.com/sillsdev/thecombine/blob/6e1588204ed54beee909b885ea46dcbc68083f61/Backend/Controllers/LiftController.cs#L168)):*

What if `_liftService.LdmlImport()` throws though? To me it seems like we should only set `LiftImported` to `true` after everything has fully completed successfully. Or is changing this behavior part of what you are needing to fix for macOS?
</blockquote></details>

In my case, the words loaded into the project, then it threw an error trying to read in the character set, so `.LiftImported` remained `false` even though the words successfully imported. So it allows me to try another import, but that fails right away, since some folder/file it tries to create already exists. I don't think that, once the words successfully import, we want to allow a second import.... unless we add a check for the imported words and, if they exists, jump to the `LdmlImport()`. But I thought that the whole reason for `.LiftImported` was because the manual folder check was unreliable.

PR #945, opened 2021-01-23
Remove reference to missing index.htm file

johnthagen:
There is no `index.htm` file generated by the user guide.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/945)

### DIFF
--- a/nginx/templates/thecombine.conf.template
+++ b/nginx/templates/thecombine.conf.template
@@ -45,9 +45,9 @@ server {
 
     # User Guide static files.
     location /docs {
-        alias       /usr/share/nginx/user_guide;
-        index      index.html index.htm;
-        expires     12h;
+        alias      /usr/share/nginx/user_guide;
+        index      index.html;
+        expires    12h;
         add_header Cache-Control "public, no-transform";
     }
 


jmgrady (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/945)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @johnthagen)

___
*[nginx/templates/thecombine.conf.template, line 49 at r1](https://reviewable.io/reviews/sillsdev/thecombine/945#-MRkQEaE84-T6FAgdrWf:-MRkQEaE84-T6FAgdrWg:bb6v03k) ([raw file](https://github.com/sillsdev/thecombine/blob/8370bb3da8cbec68106c62afd2f269c5cd3562c5/nginx/templates/thecombine.conf.template#L49)):*
> ```
>     location /docs {
>         alias       /usr/share/nginx/user_guide;
>         index      index.html index.htm;
> ```

The index statement does not suggest that these files _should_ be there but that if files with any of these names exist in the directory (going from left to right) that file should be returned when the directory is requested.  These are standard index file names so I believe that `index.htm` should be left in the list.

See the discussion of the `index` configuration item at https://docs.nginx.com/nginx/admin-guide/web-server/serving-static-content/

PR #955, opened 2021-01-29
Update frontend dependencies and move to React 17

johnthagen:
Update all frontend dependencies and migrate to React 17.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/955)


johnthagen:
Some of our dependencies still have peer dependencies on React 16, so it would be safer to close this for now.

PR #966, opened 2021-01-31
Update backup script

jmgrady:
The backup script was failing and left TheCombine shutdown.

The backup script would work properly when run manually but would not work when run from `cron`.  There were two issues that needed to be addressed:

- programs called from the script should be specified by the full path since the shell used by cron has only `/usr/bin` and `/bin` in its PATH.
- `docker-compose exec ...` assumes it is being run from an interactive session.  As a result, the call to backup the database (`docker-compose exec database mongodump ...`) would fail when called from a `cron` job.  This was addressed by adding a `-T` option when the script is running in a non-interactive session.

Additional changes include:
- adding an option to put verbose output in the syslog
- update default AWS profiles for backup
- added defaults for `combine_addl_domain_list` and `combine_cert_proxy_list` variables (in `docker_deploy/vars/config_common.yml`)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/966)


imnasnainaec (commented):
> Reviewed 2 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/966)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @jmgrady)

___
*[docker_deploy/roles/combine_backup/templates/combine-backup.j2, line 94 at r1](https://reviewable.io/reviews/sillsdev/thecombine/966#-MSY87DS9TfHf3LK3Dl2:-MSY87DT58lKxxknqMeY:b-iqgm7t) ([raw file](https://github.com/sillsdev/thecombine/blob/fc79907ee97f46ffbc3a1707d3159a60bf954760/docker_deploy/roles/combine_backup/templates/combine-backup.j2#L94)):*
> ```django
> 
> echo_verbose "4. Start up just the backend and the database."
> /usr/local/bin/aws ecr get-login-password --profile {{ aws_ecr_profile }} | docker login --username AWS --password-stdin {{ aws_ecr }}
> ```

Might `docker login`, `docker ps`, `docker cp`, etc. (in this file and others) also need a `/usr/local/bin/` prepended?
___
*[docker_deploy/roles/combine_backup/templates/combine-restore.j2, line 140 at r1](https://reviewable.io/reviews/sillsdev/thecombine/966#-MSY7ceR5i9MEKxSKKnd:-MSY7ceS87e3LwZIcxxF:bonptxo) ([raw file](https://github.com/sillsdev/thecombine/blob/fc79907ee97f46ffbc3a1707d3159a60bf954760/docker_deploy/roles/combine_backup/templates/combine-restore.j2#L140)):*
> ```django
> 
> echo verbose "5. Stop the current containers."
> docker-compose down
> ```

Do the uses of `docker-compose` in this file need to become `/usr/local/bin/docker-compose`?

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/966)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @imnasnainaec and @jmgrady)

___
*[docker_deploy/roles/combine_backup/templates/combine-backup.j2, line 94 at r1](https://reviewable.io/reviews/sillsdev/thecombine/966#-MSY87DS9TfHf3LK3Dl2:-MSYBDQE8SiC6LsTxwpC:b-ek508e) ([raw file](https://github.com/sillsdev/thecombine/blob/fc79907ee97f46ffbc3a1707d3159a60bf954760/docker_deploy/roles/combine_backup/templates/combine-backup.j2#L94)):*

Might `docker login`, `docker ps`, `docker cp`, etc. (in this file and others) also need a `/usr/local/bin/` prepended?
</blockquote></details>

Good catch.
Actually, `docker-compose` is in `/usr/local/bin` while `docker` is in `/usr/bin`.  The path for `cron` scripts is set to `/usr/bin:/bin` so the plain `docker` works.
Let me know if you would prefer the full path in the scripts.
___
*[docker_deploy/roles/combine_backup/templates/combine-restore.j2, line 140 at r1](https://reviewable.io/reviews/sillsdev/thecombine/966#-MSY7ceR5i9MEKxSKKnd:-MSYBvIDBf02uBbsmvQC:bc15tpn) ([raw file](https://github.com/sillsdev/thecombine/blob/fc79907ee97f46ffbc3a1707d3159a60bf954760/docker_deploy/roles/combine_backup/templates/combine-restore.j2#L140)):*

Do the uses of `docker-compose` in this file need to become `/usr/local/bin/docker-compose`?
</blockquote></details>

I don't think it's needed since I do not expect to run `combine-restore` in anything but an interactive setting.  In that case, `/usr/local/bin` should be on the user's path.  It would be good to have for consistency's sake so I will add it.

imnasnainaec (approved):
> Reviewed 2 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/966)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #968, opened 2021-02-01
Export updates: use sanitized project name, red error icon, update signalr

### LINKED ISSUE #941: Download should have meaningful name after unzipped.
The LIFT export zip file downloaded by the user is named in the frontend according to #800 with project name and a timestamp. However, the unzipped folder is simply `Lift` and the files within are `NewLiftFile.lift` and `NewLiftFile.lift-ranges`. These should at least have the project name (if not also a timestamp).

### LINKED ISSUE #961: Export Failed Icon should be Red
I noticed that when an export fails, a blue ! icon appears. It would probably be more visually instructive if that were colored red.

imnasnainaec:
Resolves #941
Resolves #961

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/968)


src/components/ProjectExport/DownloadButton.tsx
```diff
@@ -90,7 +99,7 @@ export default function DownloadButton(props: DownloadButtonProps) {
           <IconButton
             tabIndex={-1}
             onClick={iconFunction()}
-            color={props.colorSecondary ? "secondary" : "primary"}
+            style={{ color: iconColor() }}
```
imnasnainaec, line 102: `IconButton`'s `color` prop sadly doesn't recognize `"error"` from the theme, so I had to fall back to manual styling.

jasonleenaylor (approved):
> Reviewed 2 of 3 files at r1, 2 of 5 files at r2, 2 of 3 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/968)* status: all files reviewed, 1 unresolved discussion

PR #979, opened 2021-02-03
Switch from material-table to @material-table/core; Update to react v17.

### LINKED ISSUE #853: [ReviewEntries] Any change to an an entry resets sorting by a column
Reproduce issue:
- In review entries, click on one of the column headers to sort by that column. 
  - Two ways to confirm the sort is working: (1) there should be an arrow visible by the header; (2) the rows should be sorted accordingly.
- Make any change to one of the entries (delete, record audio, or edit and save)
  - When the update completes, the column sort will be cleared, returning to default sort

Note: Entry updates do *not* clear any filters, which is the desired behavior--sorts should also maintain through entry updates.

imnasnainaec:
Switching to `@material-table/core` (a maintained fork of `material-table`) and updating `react` to v.17 seems to fix the ReviewEntries performance issue.

However, a number of our other dependencies may run into trouble with react 17:
Screen Shot 2021-02-03 at 17 33 54

Also resolves #853 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/979)


jmgrady (approved):
I checked out the project and ran `ncu -u` (see https://reactgo.com/update-npm-packages-latest/) and I see that there are a couple of packages that get updated to later versions that are in the packages.json in this PR.  I will post a comparison in Slack (Reviewable does not seem to support uploading an HTML document or Zip archive).

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/979#-:-MSh-CRU0LJP4eS-I4nw:bx2cjyc)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

johnthagen:
@imnasnainaec For what it's worth, I also switched to `@material-table/core` for some other projects I work on, so I think this was a good idea.

I've worked on getting some fixes into it recently: https://github.com/material-table-core/core/issues/89

PR #989, opened 2021-02-05
Remove unused MergeWords MergedBy and Time fields

### LINKED ISSUE #965: Consider removing several MergeWords fields
These fields in `MergeWords` are currently unused and seem redundant:

```cs
    public class MergeWords
    {
        public string MergedBy { get; set; }
        public string Time { get; set; }
    }
```

- `MergedBy`: Should we simply use the currently logged in `User.Name`/`User.Username` if this is every needed in the future?
- `Time`: The backend can query for the current time itself. This reduces risk of time being configured differently in the frontend and being inconsistent with the rest of the backend-generated timestamps.

johnthagen:
Closes #965

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/989)


johnthagen:
Unable to perform final tests until #990 is resolved.

imnasnainaec (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/989)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnthagen)

jasonleenaylor:
___
*[src/types/word.tsx, line 63 at r2](https://reviewable.io/reviews/sillsdev/thecombine/989#-MSnkoSY2JHZjPY8vGN9:-MSnkoSZ9MzC_22bPoiS:b-p7us3k) ([raw file](https://github.com/sillsdev/thecombine/blob/eacef3ba710670d8af316e3e9b61e4aaf7e5a008/src/types/word.tsx#L63)):*
> ```
> MergeWords
> ```
Post merge review: I see this change, but I don't see anything else that matches it. Is this even being used?

johnthagen:
@jasonleenaylor No it is not being used. The reason is that we are manually creating the objects that are actually sent to the database (they use UpperCamelCase rather than lowerCamelCase). They can be removed.

I wasn't sure what the policy we were following was (e.g. do we keep all datatypes the backend exposes even if we don't use them (yet)).

Happy to remove that time if you want.

PR #1026, opened 2021-02-15
Maintenance scripts for TheCombine

jmgrady:
Broaden the scope of the `combine_backup` role to include a number of maintenance scripts.  This includes the following changes:

1. Rename the `combine_backup` role to `combine_maintenance`
2. Add the following maintenance scripts:
   - add an existing user to a project as an admin or general user
   - make an existing user a site admin
   - delete a project and its associated data
3. Add playbook to install the general `admin` user
4. Create a role to install a specific version of TheCombine - this is useful for a NUC or other target machine that is not maintained by the CI/CD processes
5. Create a role to install the general admin user

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1026)


jmgrady:
Will re-write to not require separate Database container and have to split the work between a Python script on the host and a Javascript in the container.

PR #1047, opened 2021-02-26
Document use of LiftSense.Id for GUID

### LINKED ISSUE #1015: Document Combine's use of Lift Sense.Id instead of Sense.Guid
Should document why we use Lift `Id` rather than `Guid` field on import:

https://github.com/sillsdev/TheCombine/blob/6b34e2f0e64db5d0aa4f5ec5006a75109aff1de2/Backend/Services/LiftApiServices.cs#L592

johnthagen:
Closes #1015 

`LiftSense` has a `.Guid` field. Document why The Combine imports the GUID for a `Sense` from from the `.Id` field rather than the `.Guid` field.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1047)

### DIFF
--- a/Backend/Services/LiftApiServices.cs
+++ b/Backend/Services/LiftApiServices.cs
@@ -590,6 +590,8 @@ namespace BackendFramework.Services
                     {
                         SemanticDomains = new List<SemanticDomain>(),
                         Glosses = new List<Gloss>(),
+                        // FieldWorks uses LiftSense.Id for the GUID field, rather than LiftSense.Guid,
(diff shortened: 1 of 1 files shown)


imnasnainaec (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1047)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

PR #1070, opened 2021-03-08
Fix docker install handler

jmgrady:
The handlers for the `docker_install` role included a task to restart the `logrotate` service when the configuration to rotate the docker log files is created or modified.  This causes an error in Ubuntu 18.04 because `logrotate` is scheduled as a `cron` job instead of a `systemd` service.  In Ubuntu 20.04, it is ineffective since the `logrotate` service is a `oneshot` that is triggered by a `systemd` timer; restarting it is not necessary.

Note to reviewers: In addition to the issue above, the service name in the handler file is incorrectly named `rotatelog`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1070)

### DIFF
--- a/deploy/roles/docker_install/handlers/main.yml
+++ b/deploy/roles/docker_install/handlers/main.yml
@@ -12,7 +12,7 @@
     name: rsyslog
     state: restarted
 
-- name: restart rotatelog
+- name: restart logrotate
   service:
-    name: rotatelog
+    name: logrotate
     state: restarted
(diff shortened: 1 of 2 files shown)


imnasnainaec (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1070)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

Issue #1078, opened 2021-03-11
Audio copying during merge inconsistent.


imnasnainaec:
When a sense is moved from WordA to WordB, the audio should only move if WordA has no more senses and is to be deleted (see #1072). There is a check in place in `WordApiServices.cs > Merge()`, but it only sometimes works, due to the current implementation of merging--it may be that not enough information is sent to the backend after the first merge in a set of merges, when other related merges are updated.

PR #1079, opened 2021-03-11
Improve merge audio handling.

### LINKED ISSUE #1072: Moving a sense from wordA to wordB also moves all the audio.
- Create entry with vernacular "WordA", 2 senses, and a recording.
- Create another entry with  vernacular "WordB" (must be the same or similar vernacular to the first).
- Go to Data Cleanup > Merge Duplicates.
- Move one of the two senses of WordA to WordB.
- Go to Data Cleanup > Review Entries, and the audio that was on WordA will now be on WordB.

Desired behavior: The audio should be still on WordA, not moved or copied. (The audio should be moved if and only if it is the last sense.)

imnasnainaec:
Resolves #1072, mostly (words don't lose their audio anymore).
#1078 still open, but at least prevents audio duplication on a word.
Also, match MergeWords types between backend and frontend.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1079)

### DIFF
--- a/Backend.Tests/Controllers/WordControllerTests.cs
+++ b/Backend.Tests/Controllers/WordControllerTests.cs
@@ -188,7 +188,7 @@ namespace Backend.Tests.Controllers
             var mergeObject = new MergeWords
             {
                 Parent = thisWord,
-                ChildrenWords = new List<MergeSourceWord>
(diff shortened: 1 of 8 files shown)


imnasnainaec:
Subsumed in #1081

PR #1098, opened 2021-04-06
Refactor for kubernetes

jmgrady:
Refactor the current docker configuration to support moving to Kubernetes (k8s):
1. Change the container restart policy to `always`; k8s does not support `unless-stopped`;
2. Remove volume bound to the host file system.  This involves generating the runtime configuration from environment variables when the container starts rather than creating the file once at installation.

The following improvements are also included:
1. Update the NUC home page so that the JavaScript is only updating the Access Point name in the instructions;
2. Update the maintenance script installation process (better cleanup, fix permissions)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1098)


imnasnainaec (commented):
> Reviewed 6 of 9 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1098)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[nginx/init/20_combine_runtime_config.sh, line 26 at r2](https://reviewable.io/reviews/sillsdev/thecombine/1098#-MXgvZn-EiyCuFR5OsJa:-MXgvZn-EiyCuFR5OsJb:b-i1x9y4) ([raw file](https://github.com/sillsdev/thecombine/blob/7964dc6b27173d4dcd69159cb1de72aeee02cc5f/nginx/init/20_combine_runtime_config.sh#L26)):*
> ```Shell
> 
> quote_value() {
>   if [[ $1 =~ ^[+-]?[0-9]+([.][0-9]+)?$ ]] ; then
> ```

`^[+-]?[0-9]+([.][0-9]+)?$` seems to require a digit to occur before the `.`, but is that necessary? It also doesn't account for various https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_literals

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1098)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec and @jmgrady)

___
*[nginx/init/20_combine_runtime_config.sh, line 26 at r2](https://reviewable.io/reviews/sillsdev/thecombine/1098#-MXgvZn-EiyCuFR5OsJa:-MXgyltG8jCgBsMIxQSv:b-2913un) ([raw file](https://github.com/sillsdev/thecombine/blob/7964dc6b27173d4dcd69159cb1de72aeee02cc5f/nginx/init/20_combine_runtime_config.sh#L26)):*

`^[+-]?[0-9]+([.][0-9]+)?$` seems to require a digit to occur before the `.`, but is that necessary? It also doesn't account for various https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_literals
</blockquote></details>

Good point.  According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#floating-point_literals, it looks like a digit is required before the decimal point.  Also, from your reference I realized that the test for a boolean should NOT be case insensitive. (fixed)

My intent was not to cover all numeric literals - perhaps it should have been.  I have expanded the comment to explain what is supported.  Note that currently config.js only has string (quoted) and boolean values.  It seemed like a good idea to add support for numeric values in case they were needed but my approach was too simplistic.

imnasnainaec (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1098)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #1113, opened 2021-04-16
Python code improvements

jmgrady:
Python code improvements from John's comments from PR #1091.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1113)

PR #1114, opened 2021-04-16
Drop Python 3.6 support

johnthagen:
- Drop support for Python 3.6
  - Python 3.6 support was kept solely because it is the default Python version on Ubuntu 18.04. We've since migrate to Ubuntu 20.04+
- Enable `mypy` strict mode to get additional type checking

Note that [`__from__ future import annotations`](https://www.python.org/dev/peps/pep-0563/) is needed in some cases for generics that are supported in Python 3.9+ but not in Python 3.7+. What this does is defer evaluation of the type annotations so that they aren't actually run at run time (which is not needed as Python does not do runtime type checking).

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,15 +281,11 @@ To run all Python linting steps:
 (venv) $ tox
 ```
 
-To upgrade all pinned dependencies, run the following command under Python 3.6 so the requirements are
-backwards-compatible.
+To upgrade all pinned dependencies:
 
 ```bash
 (venv) $ python -m piptools compile --upgrade dev-requirements.in
 ```
-
-Then manually remove `dataclasses==` line from `dev-requirements.txt`. This is to work around a pinning issue with
-supporting Python 3.6 and 3.7+.
 
 ## Amazon Web Services
 
--- a/deploy/bin/clean_repo.py
+++ b/deploy/bin/clean_repo.py
@@ -56,7 +56,7 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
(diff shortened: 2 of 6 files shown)


johnthagen:
The current `type-check` CI failure is relegated to https://github.com/sillsdev/TheCombine/pull/1113#issuecomment-821205289

```
  deploy/roles/combine_maintenance/files/combine_restore.py:80: error: No overload variant of "__setitem__" of "list" matches argument types "int", "Tuple[Any, str]"
  deploy/roles/combine_maintenance/files/combine_restore.py:80: note: Possible overload variants:
  deploy/roles/combine_maintenance/files/combine_restore.py:80: note:     def __setitem__(self, int, str) -> None
  deploy/roles/combine_maintenance/files/combine_restore.py:80: note:     def __setitem__(self, slice, Iterable[str]) -> None
  Found 1 error in 1 file (checked 12 source files)
  ERROR: InvocationError for command /home/runner/work/TheCombine/TheCombine/.tox/type-check/bin/mypy deploy scripts (exited with code 1)
```

jmgrady (approved):
:lgtm:

> Reviewed 4 of 6 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1114#-:-MYQxaud5QWTfWcHMz6n:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnthagen)

PR #1141, opened 2021-05-04
Migrate to public repositories

jmgrady:
Migrate from AWS ECR to using Docker Hub public repositories.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1141)


johnthagen:
A note to be sure we're aware of. At one point Docker was planning to delete inactive images off Docker Hub after six months: https://www.itproportal.com/news/docker-could-be-set-to-delete-all-your-files-if-you-havent-accessed-them-for-a-while/

It looks like they have since *paused* the roll out of this until "mid 2021" (i.e. soon):

- https://www.docker.com/blog/docker-hub-image-retention-policy-delayed-and-subscription-updates/
- https://www.docker.com/pricing/resource-consumption-updates

Depending on our use case for Docker Hub, we may need a paid license to retain images long term depending on what happens later this year.

jmgrady:
Per John's comment, the real intent is to us a public repository from any provider.  The benefit to using a public repository is that we do not need to manage login secrets to pull the images.

Since creating this issue, I have worked with LTOps and we _only_ want public repositories for code that has been reviewed to minimize the possibility of sensitive data being released in public images.

jmgrady:
We will continue to use private repositories since we need to address the issue for test image builds anyway.

PR #1161, opened 2021-05-14
Use fully qualified version for Docker Compose configuration

johnthagen:
Revert change in #1098 that causes Docker Compose configuration to use 3.0 instead of 3.8.

See: https://docs.docker.com/compose/compose-file/compose-versioning/#v2-and-v3-declaration

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1161)


jmgrady (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1161)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @johnthagen)

___
*[deploy/roles/combine_config/templates/docker-compose.yml.j2, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1161#-M_g58v7BS0jtxQHlfWJ:-M_g58v7BS0jtxQHlfWK:b-vfwnal) ([raw file](https://github.com/sillsdev/thecombine/blob/5eabb76a1536c60c1242de52a3930a2a0307db51/deploy/roles/combine_config/templates/docker-compose.yml.j2#L1)):*
> ```Jinja
> version: "3.8"
> ```

I agree that the minor version should be specified but this should be the minimum version that is required based on what is in the docker-compose.yml, not the latest version that is currently available.  For example, if there are features from 3.3 in this file, we should specify that the file is version 3.3 so that parsers from 3.3 and later can be used.

johnthagen:
___
*[deploy/roles/combine_config/templates/docker-compose.yml.j2, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1161#-M_g58v7BS0jtxQHlfWJ:-M_gMjFX0pCbEiCvSehp:b41t86j) ([raw file](https://github.com/sillsdev/thecombine/blob/5eabb76a1536c60c1242de52a3930a2a0307db51/deploy/roles/combine_config/templates/docker-compose.yml.j2#L1)):*

I agree that the minor version should be specified but this should be the minimum version that is required based on what is in the docker-compose.yml, not the latest version that is currently available.  For example, if there are features from 3.3 in this file, we should specify that the file is version 3.3 so that parsers from 3.3 and later can be used.
</blockquote></details>

Release versions are here: https://docs.docker.com/compose/compose-file/compose-versioning/#version-3

johnthagen:
___
*[deploy/roles/combine_config/templates/docker-compose.yml.j2, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1161#-M_g58v7BS0jtxQHlfWJ:-M_gMngj84TTqANe3sGP:blzpfbm) ([raw file](https://github.com/sillsdev/thecombine/blob/5eabb76a1536c60c1242de52a3930a2a0307db51/deploy/roles/combine_config/templates/docker-compose.yml.j2#L1)):*

Release versions are here: https://docs.docker.com/compose/compose-file/compose-versioning/#version-3
</blockquote></details>

I guess I view this the other way, specify the minimum Docker/Compose version we wish to support. Since 3.8 is 2 years old, to keep things simple in the event we want to use compose config features we see in the docs, use a recent version like `3.8` that will be supported in all of the versions our developers and deployments have.

This is also how the recommendations from Docker's documentation are: https://docs.docker.com/compose/gettingstarted/#step-3-define-services-in-a-compose-file rather than trying to pin to an older spec.

That being said, I don't have a super strong preference here as this is not a huge deal, I just personally think it would be better practice to pick the oldest version we want to support rather than the newest we need, so we don't have to mince through old changelogs when something doesn't work we read in the docs.

jmgrady (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1161)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[deploy/roles/combine_config/templates/docker-compose.yml.j2, line 1 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1161#-M_g58v7BS0jtxQHlfWJ:-M_gPCP31vqWDVdmxTag:b49rajh) ([raw file](https://github.com/sillsdev/thecombine/blob/5eabb76a1536c60c1242de52a3930a2a0307db51/deploy/roles/combine_config/templates/docker-compose.yml.j2#L1)):*

I guess I view this the other way, specify the minimum Docker/Compose version we wish to support. Since 3.8 is 2 years old, to keep things simple in the event we want to use compose config features we see in the docs, use a recent version like `3.8` that will be supported in all of the versions our developers and deployments have.

This is also how the recommendations from Docker's documentation are: https://docs.docker.com/compose/gettingstarted/#step-3-define-services-in-a-compose-file rather than trying to pin to an older spec.

That being said, I don't have a super strong preference here as this is not a huge deal, I just personally think it would be better practice to pick the oldest version we want to support rather than the newest we need, so we don't have to mince through old changelogs when something doesn't work we read in the docs.
</blockquote></details>

I don't have super strong preferences either but I wanted to mention how I thought the version should be used - that is, this is the version required for this file.  I will approve this change and if we ever need to change the version, e.g. because we want to go with `kompose` and it does not support this yet, then we will update the version with the change that drives it.

PR #779, opened 2020-10-23
Disable TLS 1.0 and 1.1 and enable TLS 1.3 in NGINX configuration

johnthagen:
TLS 1.0 and 1.1 have known security concerns. All major browser support at least TLS 1.2.

This disables the default NGINX configuration portion that enables support for TLS 1.0 and 1.1, and also adds support for TLS 1.3, the new standard that is more secure and can also be more performant.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/779)

PR #1193, opened 2021-06-18
Enable HTTP/2 support in production NGINX configuration

johnthagen:
Enable HTTP/2 support in NGINX. This should increase page loading performance, especially for mobile users and users running with high latency (e.g. a user in France running against the Combine within the U.S.).

Also update NGINX to the latest version (1.21).

HTTP/2 works by using a binary format that is more efficient to transmit as well as using a single TCP/TLS connection for multiple requests. This reduces round trip times to set up TCP + TLS connections. We already enabled TLS 1.3 in #779, which eliminated reduced 1 round trip time. Using HTTP/2 will further increase network performance of remote users.

HTTP/2 was released in 2015 and all major browsers support it: https://caniuse.com/http2. Somewhere around 60% of websites use HTTP/2 currently.

Because NGINX proxies connections to the backend, this change is transparent to the backend itself. It will continue to get HTTP/1.1 connections from NGINX over the fast, Docker virtual network.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1193)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ COPY . ./
 RUN npm run build
 
 # Production environment.
-FROM nginx:1.19
+FROM nginx:1.21
 
 WORKDIR /app
 
--- a/nginx/templates/default.conf.template
+++ b/nginx/templates/default.conf.template
@@ -13,7 +13,7 @@ server {
 }
 
 server {
-    listen              443 ssl;
+    listen              443 ssl http2;
     server_name         ${SERVER_NAME};
     ssl_certificate     ${SSL_CERTIFICATE};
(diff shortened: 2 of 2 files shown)


johnthagen:
To test this in Chrome, right click and enable the Protocol column in the Network tab of the developer tools. `h2` means HTTP/2.

network-tools

imnasnainaec (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1193)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #1136, opened 2021-04-30
Enable OpenAPI schema generation on the backend

johnthagen:
Closes #309
Closes #1178 

### TODO

- [x] Audit all types (look for `| null`, etc)
- [x] Audit all `[Required]` attributes and ensure they are truly all required
- [x] Migrate all connections between the frontend and backend to the new `api` module and remove any unused `backend` bindings
- [x] Replace frontend interfaces with OpenAPI-generated one; create `new___()` functions where a constructor is desired.
- [x] Look at `userid` not camelCased (warning)
- Refactor backend `index.tsx` functions into multiple models based on controller
  - Index just re-exports all
  - [save for later pr]
- [x] Investigate `Empty operationId found for path` openapi-generator warnings
- [x] Remove all of the `<remarks> Get: v1/projects/{projectId}/merge/dups/{maxInList}/{maxLists}/{userId} </remarks>`
- [x] Evaluate if the non `200` response types are worth the maintenance
  - Removing all non-`200` response types from the Lift endpoint did not change the generated TypeScript

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1136)

PR #1202, opened 2021-06-18
Clean api source folder before generating OpenAPI frontend bindings

johnthagen:
#1136 enabled automatically generating OpenAPI bindings for the frontend. One bug in this implementation is that the generation is strictly _additive_, meaning is a model or route is ever totally removed, the old file will still exist when the API is regenerated.

This PR fixes this by always totally deleting `src/api` prior to regeneration, and as a consequence found a few spurious extra files as a consequence.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1202)


imnasnainaec (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1202)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnthagen)

PR #1329, opened 2021-08-27
Update clean_repo.py

jmgrady:
Move clean_repo.py from deploy/bin to scripts.  The usage of clean_repo.py is changed as follows:
 - only removes those images that match a patter (instead of removing all images except those that match a pattern), and;
 - can take a list of repos to clean up.

Note that the CI/CD build configurations in TeamCity must be updated to use the new script before this is merged to `master`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1329)


jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1329)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec and @jmgrady)

___
*[scripts/clean_repo.py, line 69 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1329#-Mi85E7OE_HGLtY54V3U:-MiBsyJP4RvD-u6APSPK:b-3juksu) ([raw file](https://github.com/sillsdev/thecombine/blob/8921e89245687bf72083240734541bbbbb774066/scripts/clean_repo.py#L69)):*

Are we keeping `repo` instead of changing to `repos` to avoid changes in other scripts interacting with this one?
</blockquote></details>

No.  Since `repo` is a positional argument, it is never actually entered on the command line or when called by CI/CD.  I kept the `repo` name to make the usage message more readable:
```
usage: clean_repo.py [-h] [--profile PROFILE] [--untagged]
                     [--keep KEEP [KEEP ...]]
                     [--remove RM_PATTERN [RM_PATTERN ...]] [--dry-run]
                     [--verbose]
                     repo [repo ...]
```
It did not make sense to me for it to print `repos [repos ...]`.
I tried to use the `dest` option to tell it to have a different name in the `args` structure (as I did with the `--remove` option) but that throws an error for positional arguments.

I added a comment where the parser is built to explain why the argument is singular.
___
*[scripts/clean_repo.py, line 175 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1329#-Mi84gp-B94H1VoFM-KU:-MiC0RPy5oTQ8_OFypBu:bz29oe7) ([raw file](https://github.com/sillsdev/thecombine/blob/8921e89245687bf72083240734541bbbbb774066/scripts/clean_repo.py#L175)):*

Rather than checking `if rm_pattern` in every step of the `for` loop(s), can't we just check it once upfront?
</blockquote></details>

I made some changes but I still check `if rm_pattern`.  I originally initialized `rm_pattern` to an empty string.  I check to see that it is not empty because an empty regular expression will match _any_ string!  I was going to initialize it to an unmatchable regular expression (see https://codegolf.stackexchange.com/questions/18393/shortest-unmatchable-regular-expression) so that I would not need to check for an empty string.  However, I realized that it is probably more efficient to check for 'None" first than to test for a regular expression match that is guaranteed to fail.

I changed to initialize `rm_pattern` to `None` and I check to see if it is not None instead of an empty string.

I did not see any other way to move the test outside of the loop since we have to check each tag and the user can specify `--untagged` or `--remove` or both.  I certainly welcome suggestions!  (I did move the setting of `rm_pattern` to outside the looping over the repos, though.)

imnasnainaec (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1329)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[scripts/clean_repo.py, line 175 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1329#-Mi84gp-B94H1VoFM-KU:-MiNNpQ-Eai4N11U1Bf7:br8lf22) ([raw file](https://github.com/sillsdev/thecombine/blob/8921e89245687bf72083240734541bbbbb774066/scripts/clean_repo.py#L175)):*

I made some changes but I still check `if rm_pattern`.  I originally initialized `rm_pattern` to an empty string.  I check to see that it is not empty because an empty regular expression will match _any_ string!  I was going to initialize it to an unmatchable regular expression (see https://codegolf.stackexchange.com/questions/18393/shortest-unmatchable-regular-expression) so that I would not need to check for an empty string.  However, I realized that it is probably more efficient to check for 'None" first than to test for a regular expression match that is guaranteed to fail.

I changed to initialize `rm_pattern` to `None` and I check to see if it is not None instead of an empty string.

I did not see any other way to move the test outside of the loop since we have to check each tag and the user can specify `--untagged` or `--remove` or both.  I certainly welcome suggestions!  (I did move the setting of `rm_pattern` to outside the looping over the repos, though.)
</blockquote></details>

Sounds reasonable (but now your comments have to be fixed to matched the new order).

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1329)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jmgrady)

___
*[scripts/clean_repo.py, line 175 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1329#-Mi84gp-B94H1VoFM-KU:-MiNRzuw3xj3mHjY9eNs:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/8921e89245687bf72083240734541bbbbb774066/scripts/clean_repo.py#L175)):*

Sounds reasonable (but now your comments have to be fixed to matched the new order).
</blockquote></details>

Done.

imnasnainaec (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1329)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

___
*[scripts/clean_repo.py, line 194 at r5](https://reviewable.io/reviews/sillsdev/thecombine/1329#-MiNaDwC7dZ3U4Nu75qm:-MiNaDwC7dZ3U4Nu75qn:bqo4qha) ([raw file](https://github.com/sillsdev/thecombine/blob/ab329cb88ebc444ce5be38fdb6b99de784e19ff5/scripts/clean_repo.py#L194)):*
> ```
>                             image_ids.append(f"imageDigest={image_struct['imageDigest']}")
>  ...
>                     image_ids.append(f'imageDigest={image_struct["imageDigest"]}')
> ```
It's curious to see the same formatted string twice with inverted quote types.

PR #1232, opened 2021-07-02
Refactor backend to use idiomatic C# 9 conventions

johnthagen:
This is a follow on to #1198, which bumped the C# version used in the backend to 9

C# 9 features utilized:

- [negated pattern (`is not`)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/patterns3#pattern-combinators)
   - This makes negation checks more natural to read
- [target-typed `new`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/target-typed-new)
  - This removes some redundancy when declaring types by avoiding the need to specify the type on both the right and left hand side (more type inference).

Other refactors:

- Ensured consistent naming of `private readonly` attributes
- Convert some immutable locals to `const`
- Bind some unused locals with `_` to make it clear they are unused.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1232)

PR #1372, opened 2021-09-10
Refactor Backend

johnthagen:
Refactor backend and clean up the following items:

- Remove unused imports
- Remove redundant parenthesis
- Remove unused methods
- Relax target-type new usage (introduced in #1232) to avoid cases were return value type inference was used, which is not always obvious
- Use target-type new where appropriate (on variable declarations where type is already written on the left)
- Reduce nesting / cognitive load by using early return on some methods
- Make some values `readonly` or `const` where appropriate

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1372)

### DIFF
--- a/Backend.Tests/Controllers/AudioControllerTests.cs
+++ b/Backend.Tests/Controllers/AudioControllerTests.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.IO;
+﻿using System.IO;
 using Backend.Tests.Mocks;
 using BackendFramework.Controllers;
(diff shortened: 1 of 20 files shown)


jasonleenaylor (approved):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1372)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #1388, opened 2021-09-24
Display popups to alert users of network request errors

### LINKED ISSUE #1125: Display a popup whenever the backend returns a 500 error
We've missed quite a few backend bugs that threw 500 Server Error responses because the frontend does not display this to the user. Concept:

- Strive that the backend never returns 500's during normal operation. Only unhandled exceptions should return 500s
  - If the backend returns 500 ever  during a normal operation, that condition should be handled by the backend and another status (such as 400 Bad Request) should be sent instead.
- Upgrade the frontend to intercept any 500s that come back and pop a snackbar or similar notifying the user of the error
  - Allow the user to somehow expand/copy to clipboard the bigger error. In development this will be the first stacktrace

johnthagen:
Closes #1125

Display an alert whenever the API receives a 4xx or 5xx response code, to notify the user of backend errors. The alert also shows if the client can't connect to the server.

Screenshot 2021-09-24 154536

## Reviewers

This alert should not show during normal, correct operations of the Combine. If alerts are shown, they should be noted and addressed in this PR or a follow on.

- This triggers on failed login attempts. Unclear if we want to keep that behavior or try to special case.
- #1390

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1388)


jasonleenaylor (approved):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1388)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnthagen)

PR #1394, opened 2021-10-01
Improve lift import error messages

johnthagen:
Improve error messages when malformed Lift files are imported.

When #1388 is merged, these will be displayed to the users.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1394)

### DIFF
--- a/Backend/Controllers/LiftController.cs
+++ b/Backend/Controllers/LiftController.cs
@@ -68,6 +68,7 @@ namespace BackendFramework.Controllers
             }
 
             var liftStoragePath = FileStorage.GenerateLiftImportDirPath(projectId);
+            const string invalidLiftFileMessagePrefix = "Malformed Lift file: ";
 
             // Clear out any files left by a failed import
             RobustIO.DeleteDirectoryAndContents(liftStoragePath);
@@ -75,13 +76,13 @@ namespace BackendFramework.Controllers
             var file = fileUpload.File;
             if (file is null)
             {
-                return BadRequest("Null File");
+                return BadRequest($"${invalidLiftFileMessagePrefix}Null File");
             }
 
             // Ensure file is not empty
             if (file.Length == 0)
             {
-                return BadRequest("Empty File");
+                return BadRequest($"{invalidLiftFileMessagePrefix}Empty File");
             }
 
             // Copy zip file data to a new temporary file
@@ -128,7 +129,7 @@ namespace BackendFramework.Controllers
                         // Both directories seemed important
                         if (numDirs == 2)
                         {
-                            return BadRequest("Your zip file should have one directory.");
+                            return BadRequest($"{invalidLiftFileMessagePrefix}Zip file does not have one directory.");
                         }
                         break;
                     }
@@ -136,7 +137,8 @@ namespace BackendFramework.Controllers
                 default:
                     {
                         return BadRequest(
-                            "Your zip file structure has the wrong number of directories.");
+                            $"{invalidLiftFileMessagePrefix}Your zip file structure has the wrong number of directories."
+                            );
                     }
             }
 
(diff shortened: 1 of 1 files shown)


jmgrady (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1394)* status: all files reviewed, 2 unresolved discussions (waiting on @johnthagen)

___
*[Backend/Controllers/LiftController.cs, line 71 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1394#-MkvxYCT432OFSHaXR67:-MkvxYCU3QLdYfo2ioJZ:bow5kcm) ([raw file](https://github.com/sillsdev/thecombine/blob/13909ca8f25398d80a077edf60da828b25947dd6/Backend/Controllers/LiftController.cs#L71)):*
> ```
> "Malformed Lift file: "
> ```
If the error messages are to be meaningful, they should be localized - but I suppose that is a frontend function.  This would suggest that an error code should be returned to the frontend and the frontend should translate that into a localized message.

That said, I see that previous work has the error messages that are all returned in English.
___
*[Backend/Controllers/LiftController.cs, line 140 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1394#-Mkw-8X18b9PxBQ9q1oM:-Mkw-8X18b9PxBQ9q1oN:b-gik4pj) ([raw file](https://github.com/sillsdev/thecombine/blob/13909ca8f25398d80a077edf60da828b25947dd6/Backend/Controllers/LiftController.cs#L140)):*
> ```
> Your zip file structure has the wrong number of directories
> ```
It would be more helpful to tell the user what is the right number of directories.

johnthagen:
___
*[Backend/Controllers/LiftController.cs, line 71 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1394#-MkvxYCT432OFSHaXR67:-Mkw8F8pD1P3A-GpOdwm:bj224cf) ([raw file](https://github.com/sillsdev/thecombine/blob/13909ca8f25398d80a077edf60da828b25947dd6/Backend/Controllers/LiftController.cs#L71)):*

> ```
> "Malformed Lift file: "
> ```
If the error messages are to be meaningful, they should be localized - but I suppose that is a frontend function.  This would suggest that an error code should be returned to the frontend and the frontend should translate that into a localized message.

That said, I see that previous work has the error messages that are all returned in English.
</blockquote></details>

This is an interesting point that merits further discussion.

johnthagen:
___
*[Backend/Controllers/LiftController.cs, line 140 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1394#-Mkw-8X18b9PxBQ9q1oM:-Mkw8qDP75VQkSLr4GIG:b-896fix) ([raw file](https://github.com/sillsdev/thecombine/blob/13909ca8f25398d80a077edf60da828b25947dd6/Backend/Controllers/LiftController.cs#L140)):*

> ```
> Your zip file structure has the wrong number of directories
> ```
It would be more helpful to tell the user what is the right number of directories.
</blockquote></details>

Done.

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1394)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jmgrady)

___
*[Backend/Controllers/LiftController.cs, line 140 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1394#-Mkw-8X18b9PxBQ9q1oM:-Mkw9UykEjkEjczI1vsf:byskf2g) ([raw file](https://github.com/sillsdev/thecombine/blob/13909ca8f25398d80a077edf60da828b25947dd6/Backend/Controllers/LiftController.cs#L140)):*

Done.
</blockquote></details>

The default case still says that it has the wrong number of directories.  I think that the logic could be simpler if we just check for the `__MACOSX` directory first and then:
```
if (numDirs == 1)
{
   // we're good
}
else
{
   return BadRequest($"{invalidLiftFileMessagePrefix}Zip file does not have one directory.");
}
```

jmgrady (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1394)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnthagen)

PR #1474, opened 2021-11-12
Use ? with boolean props to allow implicit declaration

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1474)


johnthagen:
___
*[src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx, line 320 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1474#-MoK9xiJ1bLAZq-3NWy_:-MoK9xiJ1bLAZq-3NWya:b-pfahi4) ([raw file](https://github.com/sillsdev/thecombine/blob/1fc35057b319a05d64d16eebfc64b1b3470652f3/src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx#L320)):*
For reference, this is the docs on this: https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true

Specifically the React team seems to recommend against this (though the double negative in their statement is confusing):

> In general, we don’t recommend not passing a value for a prop, because it can be confused with the ES6 object shorthand {foo} which is short for {foo: foo} rather than {foo: true}. This behavior is just there so that it matches the behavior of HTML.

johnthagen:
___
*[src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx, line 320 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1474#-MoK9xiJ1bLAZq-3NWy_:-MoKBRnw2FkiNkxSMiOj:b59ym0x) ([raw file](https://github.com/sillsdev/thecombine/blob/1fc35057b319a05d64d16eebfc64b1b3470652f3/src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx#L320)):*

For reference, this is the docs on this: https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true

Specifically the React team seems to recommend against this (though the double negative in their statement is confusing):

> In general, we don’t recommend not passing a value for a prop, because it can be confused with the ES6 object shorthand {foo} which is short for {foo: foo} rather than {foo: true}. This behavior is just there so that it matches the behavior of HTML.


</blockquote></details>

Here the user argues that AirBnB style guide enforces that the implicit style (what is in this PR) is used everywhere: https://stackoverflow.com/a/60817451

We could even enforce this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md

johnthagen:
___
*[src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx, line 320 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1474#-MoK9xiJ1bLAZq-3NWy_:-MoKD3OJEIqu_NjWEwSq:bbcstzq) ([raw file](https://github.com/sillsdev/thecombine/blob/1fc35057b319a05d64d16eebfc64b1b3470652f3/src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx#L320)):*

Here the user argues that AirBnB style guide enforces that the implicit style (what is in this PR) is used everywhere: https://stackoverflow.com/a/60817451

We could even enforce this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md
</blockquote></details>

In summary, I think this is fine to do if we answer these questions:

- Are we okay doing this against the advisement of the React docs
- If we *do* do this, should we enable the ESLint rule above to enforce we do it everywhere consistently?

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1474)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @johnthagen)

___
*[src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx, line 320 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1474#-MoK9xiJ1bLAZq-3NWy_:-MoKXeiu3wX8ZFgZzeyd:bgshqfu) ([raw file](https://github.com/sillsdev/thecombine/blob/3c43218f201b2d9b781db42d60908c7c76a471ec/src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx#L320)):*

In summary, I think this is fine to do if we answer these questions:

- Are we okay doing this against the advisement of the React docs
- If we *do* do this, should we enable the ESLint rule above to enforce we do it everywhere consistently?
</blockquote></details>

Enforced!

johnthagen (approved):
> Reviewed 16 of 16 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1474)* status: 16 of 17 files reviewed, all discussions resolved (waiting on @johnthagen)

johnthagen (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1474)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #1480, opened 2021-11-19
Setup qa for kubernetes

jmgrady:
This PR makes the following adjustments to our installation procedure in order to enable running the QA server on Kubernetes:

1. Setup the TLS secret so that the site can be accessed securely - the site requires HSTS so that one cannot create an exception for the self-signed certificate that was being used previously.  This also makes the QA server installation a closer copy of the production installation.
2. Split the creation of the admin user from the management of the Kubernetes resources.
3. Removed the `playbook_kube_config.yml` playbook and the matching task file; the kubernetes deployments will be updated using the `kubectl set image ...` command (rolling update) outside of Ansible.
4. Update the deployment definitions to use `RollingUpdate` strategy for the QA and Live servers; `Recreate` for the NUCs.
5. Add the Kubernetes cluster for the QA server to the set of `qa` hosts in `hosts.yml`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1480)


imnasnainaec (approved):
> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1480)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

Issue #1466, opened 2021-11-08
[DataEntryTable] Recent Entries momentarily disappear when they are edited


imnasnainaec:
To avoid user edits before an update is complete, and edited Recent Entry will temporarily disappear, to reappear once the update is complete. This can be quite disconcerting. Consider disabling (and perhaps visually greying out) entries that are updating, rather than having them vanish completely.

PR #1560, opened 2022-02-09
[DataEntryTable] Improve updating of recent entries

imnasnainaec:
Specifically:
* While a recent entry is updating, temporarily disable it instead of making it disappear (Resolves: #1466)
* After an entry is updated, remove its old id from the `defunctWordIds` array (used to determine if an entry is updating)

Also:
* Move handling of `allVerns` inside `NewEntry`, since it isn't used in any other part of `DataEntryTable`.

The first item above improves the ux.
The second and third items above appear to improve performance.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1560)

### DIFF
--- a/src/components/DataEntry/DataEntryTable/DataEntryTable.tsx
+++ b/src/components/DataEntry/DataEntryTable/DataEntryTable.tsx
@@ -45,7 +45,6 @@ interface WordAccess {
 }
 
 interface DataEntryTableState {
-  existingVerns: string[];
   existingWords: Word[];
   recentlyAddedWords: WordAccess[];
   isReady: boolean;
@@ -98,7 +97,6 @@ export class DataEntryTable extends React.Component<
   constructor(props: DataEntryTableProps & LocalizeContextProps) {
     super(props);
     this.state = {
-      existingVerns: [],
       existingWords: [],
       recentlyAddedWords: [],
       isReady: false,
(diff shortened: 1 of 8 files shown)


jasonleenaylor (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1560)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

___
*[src/components/DataEntry/DataEntryTable/DataEntryTable.tsx, line 475 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1560#-MvUZpzZ0qEfmRuJJ6N8:-MvUZpzZ0qEfmRuJJ6N9:b-l46ghl) ([raw file](https://github.com/sillsdev/thecombine/blob/4967899243d1e1089a25f6de063b8fb2b6962d4d/src/components/DataEntry/DataEntryTable/DataEntryTable.tsx#L475)):*
> ```XML
>   replaceInDisplay(oldWordId: string, word: Word, removedSenseIndex?: number) {
>     this.setState((prevState) => {
>       const recentlyAddedWords = [...prevState.recentlyAddedWords];
> ```

Oh yeah, that's the right way to do that. :smile: There may be some other suboptimal code like this around, we should try to find and fix the other places as well.



_Code quote:_
```XML
    this.setState((prevState) => {
      const recentlyAddedWords = [...prevState.recentlyAddedWords];
```

PR #1624, opened 2022-03-25
Fix Helm deployment of the certificate proxy server

jmgrady:
Address a set of small problems which prevent the cert proxy server from deploying properly:

- the global definition of `serverName` was used for both the application frontend and the combine-cert-proxy server.  This was addressed 2 ways:
   - changed the structure of the `overrides` in the configuration file to have separate override YAML for each chart, i.e.
      ```
      overrides:
         chart_name:
            override_variable_1: some_value
      ...
      ```
   - created a separate variable, `certProxyServerName`, for the combine-cert-proxy.  This variable is not in the `global` section because it does not need to be shared with subcharts
- The ingress template created an ingress resources for each of the nucs.  The template was missing the `---` document separator so a single resource was created with many duplicate fields.
- Fixed the web page content for the `nuc-proxy-server`.  Changed the configmap that has the web page content from `data` to `binaryData` to accommodate the image files.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1624)

### DIFF
--- a/deploy/helm/cert-proxy-server/templates/env-nginx-configmap.yaml
+++ b/deploy/helm/cert-proxy-server/templates/env-nginx-configmap.yaml
@@ -5,4 +5,4 @@ metadata:
   name: {{ .Values.envNginxProxy }}
   namespace: {{ .Release.Namespace }}
 data:
-  SERVER_NAME: {{ .Values.global.serverName }}
+  SERVER_NAME: {{ .Values.certProxyServerName }}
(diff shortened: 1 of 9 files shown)


johnthagen (approved):
> Reviewed 8 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1624)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #1634, opened 2022-04-15
Use CompareNetObjects for scalable comparison of types

### LINKED ISSUE #1599: Migrate backend models and tests to CompareNETObjects
We should be able to remove a significant amount of boilerplate from our Models and tests by migrating all of our custom equality logic to [`CompareNETObjects`](https://nuget.org/packages/CompareNETObjects).

In tests, use [`ShouldCompare()`](https://github.com/GregFinzer/Compare-Net-Objects/wiki/Test-Extensions#shouldcompare)

johnthagen:
Closes #1599

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1634)


johnthagen:
So, in general there are a few projects I've run into applying this at scale.

- We use `List` for many objects, but when it comes to equality, we ignore the order of the objects within in. Ideally, we should consider [`HashSet`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1?redirectedfrom=MSDN&view=net-6.0) to better represent that (if our Mongo adapter allows)
  - The problem with this is that when we compare we need to set `IgnoreCollectionOrder = true` in a bunch of places. This makes the comparisons more inefficient. When I try it at our top level `Project` I even get a Stack Overflow because it was probably trying to do some kind of combinatorics on the comparison

PR #1637, opened 2022-04-20
Fix setting of server list for the cert proxy server

jmgrady:
An error was introduced when porting the Kubernetes installation from Ansible to Helm.  The variable `.Values.combineCertProxyList` is used in two ways:
1. iterate over the servers in the list to create an ingress for each server, and
2. create an environment variable for the `combine-cert-proxy` deployment.  The deployment monitors watches for changes to secrets, if the secret is for a server in the environment variable, then the secret is pushed to AWS S3.

For the first case, the variable must be a list.  When the environment variable is set from a list, it has brackets ( [ ] ) around the list which prevents the deployment from detecting any changes.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1637)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1637)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[deploy/helm/cert-proxy-server/templates/env-cert-proxy-configmap.yaml, line 9 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1637#-N0BKIvyA4JA2fj2YUa5:-N0BKIvz9HmHp5bzAlLb:b-sd45kj) ([raw file](https://github.com/sillsdev/thecombine/blob/f7ff634fd883b2d5656aaa8f66d03ea785f92903/deploy/helm/cert-proxy-server/templates/env-cert-proxy-configmap.yaml#L9)):*
> ```YAML
> data:
>   AWS_S3_BUCKET: {{ print "s3://" .Values.global.awsS3Location "/" .Values.awsS3CertLoc | quote }}
>   CERT_PROXY_CERTIFICATES: {{ template "cert-proxy-server.toString" .Values.combineCertProxyList }}
> ```

This is the only place you're not filling a `{{ template ` with ` . }}`

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1637)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jmgrady)

___
*[deploy/helm/cert-proxy-server/templates/_helpers.tpl, line 35 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1637#-N0BJGzUBlkXa3SsDq27:-N0BL0G85SoOGM05vEvg:b-4brg48) ([raw file](https://github.com/sillsdev/thecombine/blob/f7ff634fd883b2d5656aaa8f66d03ea785f92903/deploy/helm/cert-proxy-server/templates/_helpers.tpl#L35)):*

missing final newline
</blockquote></details>

Done - and I found the setting in VS Code to make sure it doesn't happen again!

imnasnainaec (commented):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1637)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @jmgrady)

___
*[deploy/helm/cert-proxy-server/templates/env-cert-proxy-configmap.yaml, line 9 at r1](https://reviewable.io/reviews/sillsdev/thecombine/1637#-N0BKIvyA4JA2fj2YUa5:-N0BM8Wb4-lWOAVyKgPH:b-ll5734) ([raw file](https://github.com/sillsdev/thecombine/blob/f7ff634fd883b2d5656aaa8f66d03ea785f92903/deploy/helm/cert-proxy-server/templates/env-cert-proxy-configmap.yaml#L9)):*

Correct.  That is intentional.  I want the template to be more generic so that I do not have to make it specific to `.Values.combineCertProxyList`
</blockquote></details>

Indeed. It's a good move to use a template here. I'm observing that elsewhere when you call on a template, you don't say what you are putting into the template in the same line, but rather use `. }}` and define what the `.` is pulled from in a different line of the code. I wasn't sure if you might want something like that here too for consistency (or even if something like that is even possible in this situation).

jmgrady:
Problem is resolved with a cleaner solution in PR #1638.

PR #1638, opened 2022-04-21
Fix setting of server list for the cert proxy server

jmgrady:
An error was introduced when porting the Kubernetes installation from Ansible to Helm. The variable .Values.combineCertProxyList is used in two ways:

- iterate over the servers in the list to create an ingress for each server, and
- create an environment variable for the combine-cert-proxy deployment. The deployment monitors watches for changes to secrets, if the secret is for a server in the environment variable, then the secret is pushed to AWS S3.

For the first use, the variable must be a list. In the second use, when the environment variable is set from a list, it has brackets ( [ ] ) around the list which prevents the deployment from detecting any changes.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1638)

### DIFF
--- a/deploy/helm/cert-proxy-server/templates/ingress-nuc.yaml
+++ b/deploy/helm/cert-proxy-server/templates/ingress-nuc.yaml
@@ -1,4 +1,4 @@
-{{- range .Values.combineCertProxyList }}
+{{- range regexSplit "\\s+" .Values.combineCertProxyList -1 }}
 ---
 apiVersion: networking.k8s.io/v1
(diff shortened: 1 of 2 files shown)


imnasnainaec (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1638)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #1713, opened 2022-08-19
Added a useful resource

Vaishnaviamirayada:
Have added a useful link about how to install python on Linux from the scaler topics

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/1713)

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ for setting up the virtual environment must be run from the top-level directory 
 
 #### Linux Python Installation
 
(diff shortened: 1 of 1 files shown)


imnasnainaec:
@Vaishnaviamirayada Thank you for wanting to help improve our README. In this case, the link you added is for installing python on Debian/Fedora, but you added it to a sentence about installing `pip` and `venv` (where it is assumed that python is already installed).

PR #1814, opened 2023-01-12
Fix case-sensitive of Find a domain

fuencui:
“Find a domain” search field at top of the domain tree should not be case-sensitive.
Add method to convert input string, then passing correct format string into the searching method

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1814)

### DIFF
--- a/Backend/Repositories/SemanticDomainRepository.cs
+++ b/Backend/Repositories/SemanticDomainRepository.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using BackendFramework.Interfaces;
 using BackendFramework.Models;
 using MongoDB.Driver;
+using MongoDB.Bson;
 
 namespace BackendFramework.Repositories
 {
@@ -36,7 +37,7 @@ namespace BackendFramework.Repositories
         {
             var filterDef = new FilterDefinitionBuilder<SemanticDomainTreeNode>();
             var filter = filterDef.And(
-                filterDef.Eq(x => x.Name, name),
+                filterDef.Regex(x => x.Name, new BsonRegularExpression("/^" + name + "$/i")),
                 filterDef.Eq(x => x.Lang, lang));
             var domain = await _context.SemanticDomains.FindAsync(filter: filter);
             try
--- a/src/components/TreeView/TreeSearch.tsx
+++ b/src/components/TreeView/TreeSearch.tsx
@@ -103,7 +103,7 @@ export function useTreeSearch(props: TreeSearchProps): TreeSearchState {
       if (!isNaN(parseInt(input))) {
         domain = await getSemanticDomainTreeNode(input, lang);
       } else {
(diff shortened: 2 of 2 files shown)


jmgrady (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1814)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @fuencui and @jasonleenaylor)

___
*[`src/components/TreeView/TreeSearch.tsx` line 122 at r1](https://reviewable.io/reviews/sillsdev/TheCombine/1814#-NLfynA9EVvkE_Y5WjxB:-NLfynA9EVvkE_Y5WjxC:bxtnodl) ([raw file](https://github.com/sillsdev/TheCombine/blob/5fcda1aaae18bde9625a5969c53bceb7e6deffba/src/components/TreeView/TreeSearch.tsx#L122)):*
> ```XML
>     //String format: capitalize the first letter, lowercase for other letters
>     return input.charAt(0).toUpperCase() + input.slice(1).toLowerCase();
>   }
> ```

Since there may be languages in the future that have capitalized words in the middle of the domain name, it would be better to do this in the backend - make a case-insensitive regular expression query.  For example, to search for "land" we would search for records whose name field matches `/^land$/i` (The `^` and `$` would make it so it matches the entire name - I'm not sure this is what we want).

I think the first place to look is at
```
public async Task<SemanticDomainTreeNode?> GetSemanticDomainTreeNodeByName(string name, string lang)
```
in the file `Backend/Repositories/SemanticDomainRepository.cs`.



_Code quote:_
```XML
  // return string converted by input string, for backend searching
  function handleInputFormat(input: string): string {
    //String format: capitalize the first letter, lowercase for other letters
    return input.charAt(0).toUpperCase() + input.slice(1).toLowerCase();
  }
```

fuencui:
@jmgrady 
Thank you for your code review and messages. I found it very helpful. I agree it’s better to make a case-insensitive regular expression query at the backend instead of the frontend.

Filtering on string fields in MongoDB is case-sensitive without using regular expressions.
To implement an insensitive search use filterDef.Regex instead

I change online code from
public async Task<SemanticDomainTreeNode?> GetSemanticDomainTreeNodeByName(string name, string lang)

Code no longer needed :
// return string converted by input string, for backend searching
 function handleInputFormat(input: string): string {
  //String format: capitalize the first letter, lowercase for other letters
  return input.charAt(0).toUpperCase() + input.slice(1).toLowerCase();
 }

jmgrady (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1814)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jasonleenaylor)

PR #1873, opened 2023-02-03
Add Words-per-User page for admin

fuencui:
Hi,
Please take your time to review the changes. 
Add Words per User collection view page
Decoupling statistics frontend components

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1873)


jasonleenaylor (changes requested):
> Reviewed 18 of 24 files at r1, 4 of 11 files at r2, 26 of 26 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1873)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @fuencui)

___
*[`Backend/Controllers/WordController.cs` line 196 at r3](https://reviewable.io/reviews/sillsdev/TheCombine/1873#-NNlK88b2W3qC2Blh_LX:-NNlK88b2W3qC2Blh_LY:bel41j9) ([raw file](https://github.com/sillsdev/TheCombine/blob/0523b985ebd391c9d2c1b411a37ae3527d563dbc/Backend/Controllers/WordController.cs#L196)):*
> ```Smalltalk
> 
> 
>             var lastIndexOfSenses = word.Senses.Count - 1;
> ```

Fuen, as you suggested this logic seems fragile.

I recommend instead modifying the AppendContainedWordContents method. It already has the logic for finding new semantic domains and you can use it to add the user to them.

Another option is to add the user to the Semantic domain on the front end. I'm not sure which is better, I think I prefer the backend.



_Code quote:_
```Smalltalk
            var lastIndexOfSenses = word.Senses.Count - 1;
            var lastIndexOfSemanticDomains = word.Senses[lastIndexOfSenses].SemanticDomains.Count - 1;
            var userId = word.Senses[lastIndexOfSenses].SemanticDomains[lastIndexOfSemanticDomains].UserId;
            // Determine whether the userId needs to be updated
            // If the userId is not null ("") meaning there is new SemanticDomains add to list
            // then userId it needs to be updated otherwise not update userId
            if (userId == "")
            {
                var updateId = _permissionService.GetUserId(HttpContext);
                word.Senses[lastIndexOfSenses].SemanticDomains[lastIndexOfSemanticDomains].UserId = updateId;
            }
```

jasonleenaylor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1873)* status: 30 of 34 files reviewed, 2 unresolved discussions (waiting on @fuencui)

___
*[`Backend/Models/Word.cs` line 247 at r5](https://reviewable.io/reviews/sillsdev/TheCombine/1873#-NNnOnXB4XlQZfjAPwD-:-NNnOnXB4XlQZfjAPwD0:b-npyxlc) ([raw file](https://github.com/sillsdev/TheCombine/blob/b0f767bc8af26c7eafc1c30a3b855a88abadee73/Backend/Models/Word.cs#L247)):*
> ```Smalltalk
>                     // For old project UserIdProperty == null no need to update
>                     // Not sure is this correct Or necessary because the MongoDB may add userId with a value "" to all SemanticDomain model once merge
>                     // also could check  (UserIdProperty.PropertyType == typeof(string) && UserIdProperty.ToString() != "")
> ```

If MongoDB may be adding an empty string then we should test with the `string.IsNullOrEmpty()` method.

I still think we might need the set logic, I think that the word sent from the front end could have semantic domains that the user did not update during this session. 

If we import a word with a sense that has a semantic domain. (Using the LIFT import feature) and then we add a semantic domain to that word we only want the user to get credit for the new semantic domain. I don't want to rely on the most recent in case we allow users to add more than one in the front-end before updating the back end. This may already be possible in Review Entries.



_Code quote:_
```Smalltalk
                    // Not sure is this correct Or necessary because the MongoDB may add userId with a value "" to all SemanticDomain model once merge
                    // also could check  (UserIdProperty.PropertyType == typeof(string) && UserIdProperty.ToString() != "")
```

jasonleenaylor (changes requested):
> Reviewed 6 of 8 files at r6, 2 of 2 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 5 of 5 files at r10, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1873)* status: 36 of 37 files reviewed, 3 unresolved discussions (waiting on @fuencui)

___
*[`Backend/Services/StatisticsService.cs` line 68 at r9](https://reviewable.io/reviews/sillsdev/TheCombine/1873#-NOAbOL16n8o5oWiUfTE:-NOAbOL16n8o5oWiUfTF:b-rylvy7) ([raw file](https://github.com/sillsdev/TheCombine/blob/0aec632fe2bb1b1cd119957e0ea9e9686aad7ec9/Backend/Services/StatisticsService.cs#L68)):*
> ```Smalltalk
>             Dictionary<string, SemanticDomainUserCount> resUserMap = new Dictionary<string, SemanticDomainUserCount>();
> 
>             // Get all user of the project
> ```

Minor grammar error, should be 'Get all users of the project'
___
*[`Backend/Services/StatisticsService.cs` line 78 at r9](https://reviewable.io/reviews/sillsdev/TheCombine/1873#-NOAbdHG8acNsPYAokT3:-NOAbdHG8acNsPYAokT4:b-il9z3e) ([raw file](https://github.com/sillsdev/TheCombine/blob/0aec632fe2bb1b1cd119957e0ea9e9686aad7ec9/Backend/Services/StatisticsService.cs#L78)):*
> ```Smalltalk
>             }
> 
>             // unknownUser is for legacy data model without a userId under SemanticDomain model
> ```

If you delete 'model' from 'data model' in this sentence it is more clear that it is old data that has no userId, not a different data model.
___
*[`src/components/DataEntry/DataEntryTable/DataEntryTable.tsx` line 178 at r11](https://reviewable.io/reviews/sillsdev/TheCombine/1873#-NOL6kSM0Wu0eUh5wEad:-NOL6kSNChO2aaA2Ybyj:b-ipmtlf) ([raw file](https://github.com/sillsdev/TheCombine/blob/53ef04154ef01eb2193f958ffaab0278f0de1261/src/components/DataEntry/DataEntryTable/DataEntryTable.tsx#L178)):*
> ```XML
> 
>     const word = await backend.getWord(wordId);
>     //update userId to new word
> ```

I am not sure about this change. Isn't the userId set when the semantic domain is added?



_Code quote:_
```XML
    //update userId to new word
    word.senses[0].semanticDomains[0].userId = getCurrentUser()?.id;
    await backend.updateWord(word);
```

jasonleenaylor (commented):
> Reviewed 2 of 3 files at r14.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1873)* status: 37 of 38 files reviewed, all discussions resolved (waiting on @fuencui)

___
*[`src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx` line 162 at r14](https://reviewable.io/reviews/sillsdev/TheCombine/1873#-NOLKOcCEJw9Z8F8v6DU:-NOLKOcD0PWZ7tGVkXbz:br713l5) ([raw file](https://github.com/sillsdev/TheCombine/blob/de6d94205800fc35df6da3ad203d073ca9732a3d/src/components/DataEntry/DataEntryTable/NewEntry/NewEntry.tsx#L162)):*
> ```XML
> 
>   updateGlossField(newValue: string): void {
>     this.props.semanticDomain.userId = getCurrentUser()?.id;
> ```

As I mentioned in slack, consider setting this in the props when creating NewEntry from the DataEntryTable class.



_Code quote:_
```XML
this.props.semanticDomain.userId = getCurrentUser()?.id;
```

jasonleenaylor (approved):
> Reviewed 1 of 3 files at r14, 1 of 2 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1873)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fuencui)

PR #1969, opened 2023-03-20
Dependabot updates for week of 20 Mar 2023

jmgrady:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1969)


jasonleenaylor (approved):
I noticed we were on an alpha for @mui/lab which has a higher likelihood of breaking on even a minor change, so I tried to find where we were using and didn't find it. I'm approving this change, but I wonder if we should look for any unnecessary dependencies.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1969#-:-NQzduif-3Fg828kQbDX:b41yava)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #1970, opened 2023-03-20
Make data statistics available for project owner

fuencui:
Make the Data Statistics tab available to the project owner
Add projectID to the route while calling statistic API 
Fix permission checking for the project owner at StatisticsController

### DIFF
--- a/Backend/Controllers/StatisticsController.cs
+++ b/Backend/Controllers/StatisticsController.cs
@@ -12,7 +12,8 @@ namespace BackendFramework.Controllers
 {
     [Authorize]
     [Produces("application/json")]
-    [Route("v1/statistics")]
+    [Route("v1/projects/{projectId}/statistics")]
+
 
     public class StatisticsController : Controller
     {
@@ -90,7 +91,6 @@ namespace BackendFramework.Controllers
 
             return Ok(await _staService.GetLineChartRootData(projectId));
         }
-
 
 
 
--- a/Backend/Controllers/StatisticsController.cs
+++ b/Backend/Controllers/StatisticsController.cs
@@ -33,7 +33,7 @@ namespace BackendFramework.Controllers
         [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(List<SemanticDomainCount>))]
         public async Task<IActionResult> GetSemanticDomainCounts(string projectId, string lang)
         {
-            if (!await _permissionService.HasProjectPermission(HttpContext, Permission.WordEntry))
+            if (!await _permissionService.HasProjectPermission(HttpContext, Permission.Owner))
             {
                 return Forbid();
             }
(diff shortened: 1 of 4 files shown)


jmgrady (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1970)* status: all files reviewed, 1 unresolved discussion (waiting on @fuencui)

___
*[`src/components/AppBar/NavigationButtons.tsx` line 72 at r1](https://reviewable.io/reviews/sillsdev/TheCombine/1970#-NR40eMk4gCivFuHSeWE:-NR40eMk4gCivFuHSeWF:bl7u6hq) ([raw file](https://github.com/sillsdev/TheCombine/blob/5c4e3f8bd6b31e23e68e015bf9a1d275f5948ffd/src/components/AppBar/NavigationButtons.tsx#L72)):*
> ```XML
>         {t("appBar.dataCleanup")}
>       </Button>
>       {isAdminOrOwner && (
> ```

Couldn't you just call `getIsAdminOrOwner()` here instead of using a state?  If I understand correctly, the function is called every time the component rendered.  If the state is set then it forces a re-render.



_Code quote:_
```XML
{isAdminOrOwner
```

jmgrady (commented):
An additional question that I have is, what if a user is:
- a member of the project but not the owner; &
- a site admin?

I believe that, in this case, the button for Data Statistics will be displayed but the backend will say that access is forbidden.  Is that correct?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1970#-:-NR41IEx9t2HO-E7zsYb:b-w0gjrw)* status: all files reviewed, 1 unresolved discussion (waiting on @fuencui)

jmgrady (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1970)* status: all files reviewed, 1 unresolved discussion (waiting on @fuencui)

___
*[`src/components/AppBar/NavigationButtons.tsx` line 31 at r1](https://reviewable.io/reviews/sillsdev/TheCombine/1970#-NR44DcD19oJPYrOxWJc:-NR44DcE7PyCNGf04Wz4:by8m32q) ([raw file](https://github.com/sillsdev/TheCombine/blob/5c4e3f8bd6b31e23e68e015bf9a1d275f5948ffd/src/components/AppBar/NavigationButtons.tsx#L31)):*
> ```XML
>   }
>   return false;
> }
> ```

An additional question that I have is, what if a user is:

a member of the project but not the owner; &
a site admin?
I believe that, in this case, the button for Data Statistics will be displayed but the backend will say that access is forbidden. Is that correct?

(I added this as a general comment but it seems it needs to be here for something that we need to resolve.)



_Code quote:_
```XML
export async function getIsAdminOrOwner(): Promise<boolean> {
  const user = LocalStorage.getCurrentUser();
  if (user?.isAdmin) {
    return true;
  } else {
    const projectId = LocalStorage.getProjectId();
    const userRoleID = user?.projectRoles[projectId];
    if (userRoleID) {
      return backend.getUserRole(userRoleID).then((role) => {
        return role.permissions.includes(Permission.Owner);
      });
    }
  }
  return false;
}
```

fuencui:
Thank you for the question.  I believe the answer is that the admin always has permission.  The 'HasProjectPermission' method at PermissionService always returns true for the admin.  Additionally, there is a comment that states administrators implicitly possess all permissions.

To sum up, the admin always has access to the statistics page (button), and the backend won't return forbidden.  The project owner also has access to the statistics page.  However, other non-admin users will not be able to see the button.

jmgrady (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/1970)* status: all files reviewed, 1 unresolved discussion (waiting on @fuencui)

PR #2074, opened 2023-04-24
Fix minor existing data drawer bugs

### LINKED ISSUE #2072: [DataEntry] For narrow window, existing entries drawer covered by top navigation bar
**Describe the bug**

In DataEntry, when the window is wide enough, a list of existing entries in the current semantic domain shows up on the right. When the window is not wide enough, there is instead an icon below the NewEntry vernacular field that (when clicked) opens a drawer on the left with the list. However, the drawer pops up underneath the site's top navigation bar, so the top of the list is cut off.

### LINKED ISSUE #2073: [DataEntry] For narrow window, existing entry button shows up even when there are none
**Describe the bug**

In DataEntry, when the window is wide enough, a list of existing entries in the current semantic domain shows up on the right. When the window is not wide enough, there is instead an icon below the NewEntry vernacular field that (when clicked) opens a drawer on the left with the list. This button should not be available when there are no existing entries, but it is (and does nothing when clicked).

imnasnainaec:
Fixes #2072 
Fixes #2073

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2074)

### DIFF
--- a/src/components/DataEntry/DataEntryComponent.tsx
+++ b/src/components/DataEntry/DataEntryComponent.tsx
@@ -135,7 +135,9 @@ export default class DataEntryComponent extends React.Component<
               treeIsOpen={this.props.treeIsOpen}
               openTree={this.props.openTree}
               showExistingData={() => this.toggleDrawer(true)}
-              isSmallScreen={this.state.isSmallScreen}
+              hasDrawerButton={
+                this.state.isSmallScreen && this.state.domainWords.length > 0
+              }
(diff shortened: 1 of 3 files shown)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2074)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

Issue #1385, opened 2021-09-24
Re-evaluate usage of eslint-disable-next-line react-hooks/exhaustive-deps


imnasnainaec:
And add explicit notes anywhere we keep using it.

As @johnthagen has stated with respect to a specific instance:
> Why is this suppression needed? There seems to be some discussion back and forth about if this is generally a code smell or not.
> facebook/create-react-app#6880
> Perhaps we should at least add a note about why this needed to be suppressed?

PR #2121, opened 2023-05-05
Export fix conflicts and timing

imnasnainaec:
Follow-up to #2115
Progress toward #1385

### DIFF
--- a/Backend/Controllers/LiftController.cs
+++ b/Backend/Controllers/LiftController.cs
@@ -257,8 +257,9 @@ namespace BackendFramework.Controllers
             }
             catch
             {
-                await _notifyService.Clients.All.SendAsync(CombineHub.ExportFailed, userId);
                 _logger.LogError("Error exporting project {ProjectId}", projectId);
+                _liftService.DeleteExport(userId);
+                await _notifyService.Clients.All.SendAsync(CombineHub.ExportFailed, userId);
                 throw;
             }
         }
--- a/Backend/Services/LiftService.cs
+++ b/Backend/Services/LiftService.cs
@@ -144,7 +144,7 @@ namespace BackendFramework.Services
         public bool DeleteExport(string userId)
         {
             var removeSuccessful = _liftExports.Remove(userId, out var filePath);
-            if (removeSuccessful && filePath is not null)
+            if (removeSuccessful && filePath is not null && filePath != InProgress)
             {
                 File.Delete(filePath);
             }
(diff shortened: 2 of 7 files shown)


jmgrady (approved):
> Reviewed 6 of 7 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2121)* status: all files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`src/components/ProjectExport/DownloadButton.tsx` line 47 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/2121#-NUvnJeyA5c1SDF5YNAK:-NUvnJeyA5c1SDF5YNAL:bw0tq0y) ([raw file](https://github.com/sillsdev/TheCombine/blob/3815d1eadf6929319b80cb74933712deac51fca0/src/components/ProjectExport/DownloadButton.tsx#L47)):*
> ```XML
>   const reset = useCallback(async () => {
>     await dispatch(asyncResetExport());
>   }, [dispatch]);
> ```

Even though the backend deletes by user id instead of project id, why remove the information?  It seems reasonable that in the future we would want to use the project id, for example, to allow multiple, simultaneous downloads.

(also for changes to `src\backend\index.ts`)



_Code quote:_
```XML
    await dispatch(asyncResetExport());
  }, [dispatch]);
```

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2121)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

___
*[`src/components/ProjectExport/DownloadButton.tsx` line 47 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/2121#-NUvnJeyA5c1SDF5YNAK:-NUvtGsHA8bxVlUhMUUm:bkk82p2) ([raw file](https://github.com/sillsdev/TheCombine/blob/3815d1eadf6929319b80cb74933712deac51fca0/src/components/ProjectExport/DownloadButton.tsx#L47)):*

Even though the backend deletes by user id instead of project id, why remove the information?  It seems reasonable that in the future we would want to use the project id, for example, to allow multiple, simultaneous downloads.

(also for changes to `src\backend\index.ts`)
</blockquote></details>

We may want to, true. As things stand, this creates a bug, because users can (and should be able to*) trigger that function when no project is opened, which causes a 404 error because of the empty string that gets used for the project it. I'm inclined to more explicitly avoid that as long as the export status is so strongly user-bound. (And I think that decision was made at least in part to prevent a single user from overloading the backend with simultaneous big project exports.)

*we allow a user's export/download to continue for a project when they switch projects, or return to the project screen and thus have no project active.

PR #2138, opened 2023-05-09
Add 6 semantic domain languages

### LINKED ISSUE #1610: Top Toolbar in French covers titles in main pane.
**Describe the bug**

In the French Interface, "Nettoyage des données" is the equivalent for "data cleanup". This localisation takes 3 lines to display, and stretches the top menu bar down. 

**To Reproduce**

1. Set Browser to French. Zoom 100%
2. Open a project.
3. Try to read the top titles.
image

**Expected behavior**
I would expect content "under" the toolbar to move down instead of getting covered.

**Environment:**

- OS: Windows Desktop
- Browser Version: Chrome 99
- Zoom/Scaling at 100%, Resolution 1920x1080

imnasnainaec:
Also resolves #1610

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2138)


jmgrady (changes requested):
> Reviewed 14 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2138)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`deploy/scripts/sem_dom_import.py` line 283 at r1](https://reviewable.io/reviews/sillsdev/TheCombine/2138#-NV4i3m888UbWzDrH2nl:-NV4i3m888UbWzDrH2nm:b-1oi1ss) ([raw file](https://github.com/sillsdev/TheCombine/blob/edcbb020fd512e9312b2634bb5a8d5d44f984138/deploy/scripts/sem_dom_import.py#L283)):*
> ```Python
>         root = tree.getroot()
>         if len(root) == 1:
>             root = root[0]
> ```

The way this is written, it will accept both
```
<Lists date="05/09/2023 9:51:25 -07:00">
  <List owner="LangProject" field="SemanticDomainList" itemClass="CmSemanticDomain">
...
```
and
```
<List owner="LangProject" field="SemanticDomainList" itemClass="CmSemanticDomain">
```
as the document roots.

However, an XML document should have a single, well-defined structure and I don't know of a way to create an XML schema that will support both of these structures (caveat: I am no expert on the XML Schema Definition language).

I believe the script should be updated to use the new document structure.  If I had my druthers, I would iterate over all the children of the `<Lists>` element and process those that have the correct `field` and/or `itemClass` attributes.  This would decrease the scripts dependency on the process for creating the XML file, that is, it would be possible to have additional content.

A simpler solution would be to verify that:
 - the root element has a length of 1 (as it does here)
 - the child element has the correct `field` and/or `itemClass` attributes
If it fails either of these tests, the script should log an error and exit with a non-zero return code.



_Code quote:_
```Python
        if len(root) == 1:
            root = root[0]
```

jmgrady (commented):
> Reviewed 1 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2138)* status: all files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2138)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[`deploy/scripts/sem_dom_import.py` line 283 at r1](https://reviewable.io/reviews/sillsdev/TheCombine/2138#-NV4i3m888UbWzDrH2nl:-NV4ufHd3hu_9Ajw_kZ1:bvnrpcm) ([raw file](https://github.com/sillsdev/TheCombine/blob/edcbb020fd512e9312b2634bb5a8d5d44f984138/deploy/scripts/sem_dom_import.py#L283)):*

However, since the XML source files are part of our source tree and should be reviewed before they are committed to the repo, the "simpler" solution is more than sufficient and we may not need to check the attributes.
</blockquote></details>

updated

PR #2143, opened 2023-05-11
convert audio to wav during a lift export

hahn-kev:
This change makes it so that all audio files exported have their container changed to `wav`, currently files recorded in the browser are saved in a `webm` format with an encoding that FLEx doesn't support when in a `webm`, but if we convert it to a `wav` then it'll work just fine and preserve any high quality audio.

All tests pass, though I had to make an adjustment as the audio file names will be `wav` files instead of `mp3`. An alternative would be to only convert `webm` files to `wav` and leave everything else as is, however we may run into more issues in the future with other file formats. We know that `wav` works so it makes sense for now to just keep this change simple in my opinion. If there's a reason that we don't want to convert all files to `wav` let me know and we can adjust it.

In order for the unit tests to pass (which run outside of docker), I had to add a new step `backend.yml` to install `ffmpeg` before we run the unit tests. That file may need to be merged into master before it effects the checks in this PR, but I'm not sure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2143)

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -41,7 +41,7 @@ jobs:
         with:
           dotnet-version: ${{ matrix.dotnet }}
       - name: Install ffmpeg
-        uses: FedericoCarboni/setup-ffmpeg@0772ab98049eccf3c09d0a23208a5f1f8045a6de # v2.0.0
+        uses: FedericoCarboni/setup-ffmpeg@v2 # v2.0.0
 
       # Coverage.
       - name: run coverage tests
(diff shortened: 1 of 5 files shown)


imnasnainaec:
@hahn-kev Hi, Kevin! I think we'd prefer to only convert webm files to wav on export in order to preserve the format of any imported media files.

hahn-kev:
Thanks for the feedback @imnasnainaec I've updated the code to only convert `.webm` files to `.wav` everything else will be preserved. This does mean that out test cases aren't actually executing this code branch anymore, so we might want to add a test case with some `.webm` files, but I'm not familiar enough with this project to do that correctly.

imnasnainaec:
Replaced by #2176

Issue #708, opened 2020-09-08
[SiteSettings] Allow site admins to change user roles on a project


imnasnainaec:
Currently, UserManagement in SiteSettings only allows site admins to delete other users.

It would be nice to be able to 
- [ ] add a user to a project; (#2372)
- [x] remove a user from a project or change their role. (#2370)

This feature should be presented through the project management component.

(May be able to re-use the project settings component for users)

imnasnainaec:
Related: https://github.com/sillsdev/TheCombine/issues/544

PR #2336, opened 2023-06-29
Roles

### LINKED ISSUE #1983: Permission definitions include roles
The current user permissions are defined in `src/api/models/permission.ts` as:
```
export enum Permission {
  WordEntry = "WordEntry",
  MergeAndReviewEntries = "MergeAndReviewEntries",
  ImportExport = "ImportExport",
  DeleteEditSettingsAndUsers = "DeleteEditSettingsAndUsers",
  Owner = "Owner",
}
```
However, `Owner` is a role, not a permission.  It would be better if we removed `Owner` and add the permissions that are currently restricted to an `Owner`, e.g. `ViewStatistics`.

imnasnainaec:
Resolves #1983

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)


imnasnainaec:
Corresponding necessary database update here: https://docs.google.com/document/d/1L9WjBZqBNVfxk5KiIQo5Fzjsetv4JuwGU0wuYKDRtWo

imnasnainaec (commented):
@jmgrady "Another option" implemented.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-:-NZkejv77KheEst9IEDJ:byek9or)* status: 0 of 47 files reviewed, all discussions resolved (waiting on @imnasnainaec)

jmgrady (commented):
> Reviewed 5 of 31 files at r1, 10 of 30 files at r2, 1 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 14 of 47 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`README.md` line 863 at r5](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-NZkkYfP0KOBVTEZK_vJ:-NZkkYfP0KOBVTEZK_vK:bxqo8ri) ([raw file](https://github.com/sillsdev/TheCombine/blob/a806aa7101e63a8deee0e834c67e7bd8ca0f1d6c/README.md#L863)):*
> ```Markdown
>    `Editor` role, or the `--admin` option to instead add the user with `Administrator` role.
> 
> #### Backup _TheCombine_
> ```

I realize that I am the one who originally put the duplicate the instructions for the `add_user_to_proj.py` Python script here but I think we should just say:

Run:

```
kubectl exec -it deployment/maintenance -- add_user_to_proj.py --help`
```
for additional options.

This decreases the likelihood of documentation errors and decreases maintenance work for the script.



_Code quote:_
```Markdown
1. The `--project` and `--user` options may be shortened to `--p` and `--u` respectively.
2. The user is added to the project with `Harvester` role. Add the `--editor` option to instead add the user with
   `Editor` role, or the `--admin` option to instead add the user with `Administrator` role.
```
___
*[`Backend/Controllers/ProjectController.cs` line 182 at r5](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-NZkqiOz2NuzJB-gnJH4:-NZkqiOz2NuzJB-gnJH5:b-7r5gjr) ([raw file](https://github.com/sillsdev/TheCombine/blob/a806aa7101e63a8deee0e834c67e7bd8ca0f1d6c/Backend/Controllers/ProjectController.cs#L182)):*
> ```Smalltalk
>         [HttpDelete("{projectId}", Name = "DeleteProject")]
>         [ProducesResponseType(StatusCodes.Status200OK)]
>         public async Task<IActionResult> DeleteProject(string projectId)
> ```

This is beyond the scope of this PR but `DeleteProject` is a misnomer and will cause problems when we add the ability to actually delete projects as has been requested.



_Code quote:_
```Smalltalk
       public async Task<IActionResult> DeleteProject(string projectId)
```

jmgrady (commented):
> Reviewed 2 of 30 files at r2, 1 of 5 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 20 of 47 files reviewed, 2 unresolved discussions (waiting on @imnasnainaec)

___
*[`Backend.Tests/Controllers/UserRoleControllerTests.cs` line 104 at r7](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-NZlS50g7kThND9z3_Um:-NZlS50g7kThND9z3_Un:b-2vgf3p) ([raw file](https://github.com/sillsdev/TheCombine/blob/cd4ff2876aedf82075fce20603723f79bdddb352/Backend.Tests/Controllers/UserRoleControllerTests.cs#L104)):*
> ```Smalltalk
> 
>             var foundPermissions = ((ObjectResult)action).Value as List<Permission>;
>             Assert.AreEqual(ProjectRole.RolePermissions(userRole.Role!), foundPermissions);
> ```

This seems to assume that the permissions are always stored in the same order.  That could be a valid assumption for now but since we treat the permissions more like a set than an ordered list, it might be best to sort both lists first.



_Code quote:_
```Smalltalk
            Assert.AreEqual(ProjectRole.RolePermissions(userRole.Role!), foundPermissions);
```

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 19 of 47 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[`Backend/Controllers/ProjectController.cs` line 182 at r5](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-NZkqiOz2NuzJB-gnJH4:-N_--5kX6sTApoo0GbHc:bqoqspq) ([raw file](https://github.com/sillsdev/TheCombine/blob/a806aa7101e63a8deee0e834c67e7bd8ca0f1d6c/Backend/Controllers/ProjectController.cs#L182)):*

This is beyond the scope of this PR but `DeleteProject` is a misnomer and will cause problems when we add the ability to actually delete projects as has been requested.
</blockquote></details>

Not a misnomer here. This function actually deletes a project, but we don't use it. Our project archiving function is defined in the frontend and just uses the project update function from this controller.

jmgrady (changes requested):
> Reviewed 4 of 31 files at r1, 3 of 30 files at r2, 2 of 5 files at r4, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 28 of 47 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`maintenance/scripts/combine_app.py` line 28 at r7](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-IoV2DkWeL35LA-Ds:-N_-IoV2DkWeL35LA-Dt:b-jxeo9b) ([raw file](https://github.com/sillsdev/TheCombine/blob/cd4ff2876aedf82075fce20603723f79bdddb352/maintenance/scripts/combine_app.py#L28)):*
> ```Python
>     # Integer value 7 is currently unused.
>     # Integer value 8 is currently unused.
>     Owner = 9
> ```

This is one of 3 places where this Enum is defined - largely because it is used by code written in 3 different languages.  However, the enum values are defined as integers in the Python maintenance script and the Backend but defined as strings in the Frontend.

Since each user can only have a single role in a project, I would like to suggest that the Roles enum be changed to have string values for the Backend, the Database, and these maintenance scripts.  This would make the results of database queries more readable, minimized Enum definition errors and would obviate the need for "gaps" in the roles sequence (and the risk that any one gap that we choose today is not big enough as roles grow).

Presumably, the value of using an integer is so that one can use '>' of '<' to determine if one role has more permission than another.  However, the only place that I have found where this is used is in the maintenance script, `add_user_to_proj.py`.  The comments say that this is to make sure than a users role is never decreased through this script.  This is not a necessary requirement of the script; in fact, it may be more useful to be able to directly set the role of a new user.  Note that this script can only be executed by users who have access to running a shell on the Kubernetes cluster; it is never called by _The Combine_ application.



_Code quote:_
```Python
class Role(Enum):
    """Define enumerated type for Combine user roles."""

    # None = 0
    # Integer value 1 is currently unused.
    Harvester = 2
    # Integer value 3 is currently unused.
    Editor = 4
    # Integer value 5 is currently unused.
    Administrator = 6
    # Integer value 7 is currently unused.
    # Integer value 8 is currently unused.
    Owner = 9
```

jmgrady (changes requested):
> Reviewed 3 of 31 files at r1, 8 of 30 files at r2, 2 of 3 files at r3, 3 of 4 files at r5, 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: all files reviewed, 2 unresolved discussions (waiting on @imnasnainaec)

___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 36 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-iu8v3YMc4OG14ivw:-N_-iu8v3YMc4OG14ivx:b26fwj8) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L36)):*
> ```XML
>   );
> 
>   const [projectId] = useState(getProjectId());
> ```

If there is no function to set the projectId I am not sure that `useState` is appropriate here.  Why not just call `getProjectId()` each time it is used? (or once at the beginning of the function).

That said, I see in the file `src/components/ProjectSettings/tests/index.test.tsx` that the test relies on a change of `projectId` to be seen and to trigger a rerender.  Is this possible in the normal operation of the application?  I did not think that it was.  It suggests that the current project id is coming from outside the component - shouldn't this be a redux state and `useAppSelector` be used instead of `useState`?



_Code quote:_
```XML
  const [projectId] = useState(getProjectId());
```
___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 80 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-kwCn-AVa7ySge7Rb:-N_-kwCn-AVa7ySge7Rc:blibrau) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L80)):*
> ```XML
> 
>   const currentIsProjOwner = userRoles[currentUser.id] === Role.Owner;
>   const currentIsProjAdminOrOwner =
> ```

The checks for if a user is an Administrator or Owner suggest that there is a missing permission type that should be checked rather than checking for roles.  (Another place this is seen is for viewing Statistics only in that case, the check is for Site Administrator.)



_Suggestion:_
```XML
...AdminOrOwner
```

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 46 of 47 files reviewed, 2 unresolved discussions (waiting on @jmgrady)

___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 80 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-kwCn-AVa7ySge7Rb:-N_04E8mEljquz4JBxT0:bvkj29f) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L80)):*

The checks for if a user is an Administrator or Owner suggest that there is a missing permission type that should be checked rather than checking for roles.  (Another place this is seen is for viewing Statistics only in that case, the check is for Site Administrator.)
</blockquote></details>

My mental model of this only has the frontend fetching permissions for the current user--when a user is managing other users, since they'll be operating in terms of roles and not permissions, it felt more natural to design the corresponding frontend operations as strictly role-based.

jmgrady (commented):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: all files reviewed, 2 unresolved discussions (waiting on @imnasnainaec)

___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 80 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-kwCn-AVa7ySge7Rb:-N_0CKkgA6NctWbV3eYC:bsasjn1) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L80)):*

My mental model of this only has the frontend fetching permissions for the current user--when a user is managing other users, since they'll be operating in terms of roles and not permissions, it felt more natural to design the corresponding frontend operations as strictly role-based.
</blockquote></details>

I still think that this is indicative of an unidentified permission or permissions.  The permission in question is does the user have permission to ModifyUserRoles (which may be different than DeleteEditSettingsAndUsers)  The other possibility is to check if the user has the DeleteEditSettingsAndUsers permission.  The backend checks that you just added will make sure an admin doesn't demote an owner, or other types of coups, correct?

Also, how does a user who is a site admin fit into this?

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 44 of 47 files reviewed, 2 unresolved discussions (waiting on @jmgrady)

___
*[`maintenance/scripts/combine_app.py` line 28 at r7](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-IoV2DkWeL35LA-Ds:-N_0RZltBIgw0eyOEAy9:bb6un1j) ([raw file](https://github.com/sillsdev/TheCombine/blob/cd4ff2876aedf82075fce20603723f79bdddb352/maintenance/scripts/combine_app.py#L28)):*

This is one of 3 places where this Enum is defined - largely because it is used by code written in 3 different languages.  However, the enum values are defined as integers in the Python maintenance script and the Backend but defined as strings in the Frontend.

Since each user can only have a single role in a project, I would like to suggest that the Roles enum be changed to have string values for the Backend, the Database, and these maintenance scripts.  This would make the results of database queries more readable, minimized Enum definition errors and would obviate the need for "gaps" in the roles sequence (and the risk that any one gap that we choose today is not big enough as roles grow).

Presumably, the value of using an integer is so that one can use '>' of '<' to determine if one role has more permission than another.  However, the only place that I have found where this is used is in the maintenance script, `add_user_to_proj.py`.  The comments say that this is to make sure than a users role is never decreased through this script.  This is not a necessary requirement of the script; in fact, it may be more useful to be able to directly set the role of a new user.  Note that this script can only be executed by users who have access to running a shell on the Kubernetes cluster; it is never called by _The Combine_ application.

</blockquote></details>

Done
___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 80 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-kwCn-AVa7ySge7Rb:-N_0RnQkAWZDBMkpftN9:b-kgj9y7) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L80)):*

I still think that this is indicative of an unidentified permission or permissions.  The permission in question is does the user have permission to ModifyUserRoles (which may be different than DeleteEditSettingsAndUsers)  The other possibility is to check if the user has the DeleteEditSettingsAndUsers permission.  The backend checks that you just added will make sure an admin doesn't demote an owner, or other types of coups, correct?

Also, how does a user who is a site admin fit into this?
</blockquote></details>

#708 is for site admins (to be implemented in site settings).

I'll look again at the permissions used here.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 43 of 47 files reviewed, 2 unresolved discussions (waiting on @jmgrady)

___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 80 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-kwCn-AVa7ySge7Rb:-N_0To8Q6XinSn5So7OG:bpktupj) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L80)):*

#708 is for site admins (to be implemented in site settings).

I'll look again at the permissions used here.
</blockquote></details>

Yes, I see what you mean about permissions. Since the component in question is only visible with permission `DeleteEditSettingsAndUsers` I just simplified the check here. If we add other roles with user-management ability, it may need a more permissiony redesign.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: 42 of 47 files reviewed, 2 unresolved discussions (waiting on @jmgrady)

___
*[`src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx` line 36 at r10](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_-iu8v3YMc4OG14ivw:-N_0UjGhCDi1IrsT4kmg:bgnps1n) ([raw file](https://github.com/sillsdev/TheCombine/blob/dfa3a4d28618726ae96a402d50521cad95b8603f/src/components/ProjectSettings/ProjectUsers/ActiveProjectUsers.tsx#L36)):*

If there is no function to set the projectId I am not sure that `useState` is appropriate here.  Why not just call `getProjectId()` each time it is used? (or once at the beginning of the function).

That said, I see in the file `src/components/ProjectSettings/tests/index.test.tsx` that the test relies on a change of `projectId` to be seen and to trigger a rerender.  Is this possible in the normal operation of the application?  I did not think that it was.  It suggests that the current project id is coming from outside the component - shouldn't this be a redux state and `useAppSelector` be used instead of `useState`?
</blockquote></details>

It is possible with the `ProjectSwitch` component.

I updated this to get the project id from its parent, which fits what the other project settings components are doing.

jmgrady (changes requested):
> Reviewed 3 of 3 files at r12, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: all files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`maintenance/scripts/add_user_to_proj.py` line 50 at r14](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_0cByDDYCOfQyKKtHV:-N_0cByDDYCOfQyKKtHW:behrb1o) ([raw file](https://github.com/sillsdev/TheCombine/blob/96279fde53b19933787a5ff2af04a568713d943c/maintenance/scripts/add_user_to_proj.py#L50)):*
> ```Python
>     )
>     return parser.parse_args()
> ```

The argument parser setup can be improved.  Currently someone could specify `--admin` _and_ `--editor`.  If you do the following the user can pick one role and it stays in sync with changes to the `Role` enum:
1. remove the add_argument calls for `--editor` and `--admin`;
2. Add the following call:
```
parser.add_argument("--role", choices=[role.value for role in Role], default=Role.Harvester.value, help=...)
```
3. Remove logic for setting req_role - can set directly to args.role.
4. Probably want to add a check for args.role == Role.Owner.value and throw an error.  The other option would be to create a function that returns all the enum values _except_ Owner starting with the example above.

One other nice thing (that I know about now) is that you can specify short, single-dash options as an equivalent for the double-dash ones simply but adding it as an additional argument, for example,
```
parser.add\_argument(
   "--project", "-p", required\=True, help\="Project(s) to be removed from TheCombine."
)
```
would allow you to set the project using `--project` or `-p`.

jmgrady (approved):
> Reviewed 1 of 1 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2336)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

___
*[`maintenance/scripts/add_user_to_proj.py` line 50 at r14](https://reviewable.io/reviews/sillsdev/TheCombine/2336#-N_0cByDDYCOfQyKKtHV:-N_579fH2NTWF_W-nvLD:b1bopy2) ([raw file](https://github.com/sillsdev/TheCombine/blob/96279fde53b19933787a5ff2af04a568713d943c/maintenance/scripts/add_user_to_proj.py#L50)):*

Updated.

From my understanding, the first letter of an argument is automatically available as a single-dash arg (if it's the only arg that begins with that letter).
</blockquote></details>

No, `--p` would work but not `-p`.  Any abbreviation of the option will work if it is not ambiguous - but requires both `-` characters.

PR #2376, opened 2023-07-14
[PermissionService] Add tests

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2376)


jasonleenaylor (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2376)* status: all files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`Backend.Tests/Services/PermissionServiceTests.cs` line 49 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/2376#-Na7cbuH1M5QbzOFhNl2:-Na7cbuI7WwZuqKwcRaS:bmd79w2) ([raw file](https://github.com/sillsdev/TheCombine/blob/feed061690038312f555bfaaa646bde82fdd1618/Backend.Tests/Services/PermissionServiceTests.cs#L49)):*
> ```Smalltalk
>         public void GetUserIdTestReturnsNonemptyId()
>         {
>             Assert.False(String.IsNullOrEmpty(_permService.GetUserId(createHttpContextWithUser(new User()))));
> ```

I think we should prefer the Constraint Model for our C# NUnit tests. It is less important for the true/false checks, but I think the more we encourage using it the more comfortable it is for other tests. The Constraint failure messages prevent confusion between expected and actual values, and the failure messages tend to be more easily understood.
Here it would be something like this: `Assert.That(()=>string.IsNullOrEmpty(_permService.GetUserId(createHttpContextWithUser(new User()))), Is.False);`

Issue #2404, opened 2023-07-19
Implement Contexts


imnasnainaec:
We have some objects that are passed down to various children via prop-drilling that might be better served in a [context](https://react.dev/learn/passing-data-deeply-with-context).

- [x] DataEntry recorder (#2405)
- [ ] DataEntry spell checker (#2413)
- [x] ReviewEntries recorder (#2405)

PR #2405, opened 2023-07-19
Add a Recorder context

imnasnainaec:
Progress on #2404

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2405)


jasonleenaylor (approved):
My only concern with this is theoretical. We will have a singleton Recorder, and I'm not sure if it could ever get into a bad state after repeated use. If it could then all recording would be broken on all pages and switching pages would not fix it.
:lgtm:
:lgtm:

> Reviewed 23 of 26 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2405#-:-N_nwSw52Ff347zm-PbO:b-4go1qq)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #2430, opened 2023-07-25
Add more font support

### LINKED ISSUE #703: Make available the fonts used by the language picker
TheCombine currently uses for everything the first available font from a few san-serif fonts: Roboto, Noto Sans, Helvetica, Arial, sans-serif. However, the language picker has the user select a font to associate with each of their vernacular and analysis languages (often defaulting to Charis SIL). We store that font selection but do not actually use it.

Some scripts more urgently require font support, as their characters are showing as tofu in the app:
- Miao / Shimenkan

----

Let's support all the (non-grey, non-legacy) fonts in https://software.sil.org/fonts/

...and/or the fonts used in Mui-Language-Picker:
https://github.com/sillsdev/mui-language-picker/blob/master/src/langPicker/fontList.js#L17

----

We should be sure to ask the Audio Project Manager team about lessons they learned when they did this.
https://github.com/sillsdev/transcriber/blob/master/src/crud/fontChoice.ts#L31
https://github.com/sillsdev/transcriber/blob/master/src/components/TranscriptionShow.tsx#L6

----

There's helpful info in this README: https://github.com/sillsdev/Transcribe/tree/master/src/portable/ReactUi/react#adding-images-fonts-and-files

----

We want to lazy load the fonts so we don't include ones that are not needed for a project.

----

[https://mui.com/material-ui/customization/typography/ says](https://mui.com/material-ui/customization/typography/):
> To self-host fonts, download the font files in ttf, woff, and/or woff2 formats and import them into your code.
> This requires that you have a plugin or loader in your build process that can handle loading ttf, woff, and woff2 files. Fonts will not be embedded within your bundle. They will be loaded from your webserver instead of a CDN.

----

https://software.sil.org/fonts/webfonts/ has instructions for self-hosting SIL web fonts

imnasnainaec:
Resolves #703

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2430)


imnasnainaec:
Replaced by #2462

PR #2462, opened 2023-08-09
Add font support

### LINKED ISSUE #703: Make available the fonts used by the language picker

imnasnainaec:
Resolves #703

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2462)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2462)* status: 4 of 50 files reviewed, 4 unresolved discussions (waiting on @imnasnainaec and @jmgrady)

___
*[`Dockerfile` line 48 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc31l087g72pirI9_yy:-Nc4It17APdzaEgppsnZ:b-896fix) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Dockerfile#L48)):*

This order should not be changed.  `frontend_builder` results will change more frequently than `user_guide_builder` results.
</blockquote></details>

Done.
___
*[`Dockerfile` line 52 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc31KaQCpNrGKJFJM7_:-Nc4IthvE0jGAu3_glUy:b-896fix) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Dockerfile#L52)):*

This order should not be changed.  `public/scripts/release.js` will change more frequently than `src/resources/tractor.png`.
</blockquote></details>

Done.
___
*[`Backend/Helper/FileStorage.cs` line 139 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3ExwbD7lUz3AnkkkX:-Nc4IPz89UCVAgiExCtD:bwf168r) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/FileStorage.cs#L139)):*

Similar to the comment below, the backend should be told where the fonts are located or it should be relative to the directory in which the backend is running.
</blockquote></details>

This mimics what we have in `Backend.Tests/Utils.cs`. If `Environment.CurrentDirectory` doesn't give the directory where the backend is running, how can I obtain that?
___
*[`Backend/Helper/FileStorage.cs` line 171 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3DGHaASUPmjLAVis6:-Nc4I34w4qwLBEGr5yIb:b-iopgfi) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/FileStorage.cs#L171)):*

This should be replaced by looking up an environment variable that specifies the folder for the backend file storage.  Assuming that it is the HOME directory makes it difficult for development because if I have multiple working directories for different features or issues, they will all be using the same directory.  In addition, the Backend should be Developer OS agnostic.
</blockquote></details>

I reverted the change here. I'm inclined to leave you suggestion to another pr.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2462)* status: 4 of 50 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[`Backend/Helper/Font.cs` line 14 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3keuo6a2KOlcwMcUJ:-Nc7olK6EkcJ05NXxTit:b-5ac76) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L14)):*

Perhaps this would be yet another good use for an environment variable.  We would not want to rebuild the application in the event that Google moves their fonts.  It may be so unlikely that it's worth taking the risk.

As a show of my ignorance, what is the source of this URL?  We have a list of fonts - is this URL embedded in the original font files that we use to generate the CSS files?


It would be better to also rename the variable to something like `fontServiceUrl` or `externalFontSourceUrl` since this should work for any font source, correct?
</blockquote></details>

This URL comes from Google's font api. It is only used for Google fonts in the case that the original SIL font files we use for CSS-generation lack a font link. It wouldn't necessary work for any font source, because their semantics allow for just tacking the font name at the end of that URL stub.
___
*[`Backend/Helper/Font.cs` line 46 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3n1VT3T80jDGUmFjM:-Nc7odkxDK01_iRFxHco:be43081) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L46)):*

Should these `var`s be `let`s?
</blockquote></details>

No, that is the TS preference: https://github.com/sillsdev/TheCombine/blob/master/docs/style_guide/ts_style_guide.md#const-vs-let-vs-var

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2462)* status: 4 of 50 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`Backend/Helper/Font.cs` line 46 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3n1VT3T80jDGUmFjM:-Nc7uz77-uAEaKVvVnjp:bop4glf) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L46)):*

No, that is the TS preference: https://github.com/sillsdev/TheCombine/blob/master/docs/style_guide/ts_style_guide.md#const-vs-let-vs-var
</blockquote></details>

Actually, the way I read that document, `let` is preferred over `var` because `var` is a global variable.  Am I missing something?

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2462)* status: 4 of 50 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[`Backend/Helper/Font.cs` line 14 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3keuo6a2KOlcwMcUJ:-Nc889kfCEyXZD8hJy1l:b-1vmt9i) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L14)):*

This URL comes from Google's font api. It is only used for Google fonts in the case that the original SIL font files we use for CSS-generation lack a font link. It wouldn't necessary work for any font source, because their semantics allow for just tacking the font name at the end of that URL stub.
</blockquote></details>

Per our discussion, moving the css fetching to the frontend.
___
*[`Backend/Helper/Font.cs` line 27 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3mWuf-BEOlAtBNLeq:-Nc88FFF84hrQ0VsryhW:b-1vmt9i) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L27)):*

Again, I think it would be better to generalize references to Google.
</blockquote></details>

Per our discussion, moving the css fetching to the frontend.
___
*[`Backend/Helper/Font.cs` line 46 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3n1VT3T80jDGUmFjM:-Nc85E1o-Sw0ySjnZ8p2:b3ltyhy) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L46)):*

Actually, the way I read that document, `let` is preferred over `var` because `var` is a global variable.  Am I missing something?
</blockquote></details>

You read the document correct, but the document is about TypeScipt and this is a C# file. :smiley:

jmgrady (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2462)* status: 4 of 50 files reviewed, 1 unresolved discussion (waiting on @imnasnainaec)

___
*[`Backend/Helper/Font.cs` line 46 at r4](https://reviewable.io/reviews/sillsdev/TheCombine/2462#-Nc3n1VT3T80jDGUmFjM:-Nc8A6SLAtObPDZxPQpT:b-mli7ao) ([raw file](https://github.com/sillsdev/TheCombine/blob/b00b792a61b6d1af979d71fbb7cb6e666faef8ca/Backend/Helper/Font.cs#L46)):*

You read the document correct, but the document is about TypeScipt and this is a C# file. :smiley:
</blockquote></details>

🤦 Clearly I was not paying attention!  Thanks.

imnasnainaec:
Replaced by #2493

PR #2573, opened 2023-09-13
[ReviewEntries] Prevent duplicate semantic domain on a sense

### LINKED ISSUE #2572: [ReviewEntries] Can add duplicate semantic domain, but delete one and they all go
**Describe the bug**

In Review Entries, when you edit a row, you can add semantic domains to a sense. However, you are allowed to add a domain that's already there, and then if you try to delete a duplicate all tags for that semantic domain are deleted.

**Expected behavior**

You shouldn't be able to add a duplicate semantic domain in the first place. ~~If adding the same number with different name (e.g., in a different language), the new one should probably replace the old one.~~ (Never mind the latter part--that would override useful data like `.userId` and `.created`. The user should have to delete it to re-add it.)

imnasnainaec:
Fixes #2572

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2573)


jmgrady (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2573)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #2727, opened 2023-10-19
Define a common repository interface

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2727)


imnasnainaec:
We discussed and decided not to make this change.

PR #2746, opened 2023-10-26
[NewEntry] Improve dialog UI

### LINKED ISSUE #1342: [NewEntry] Improve Entry and Sense dialogs
Possible improvements:
- [ ] Add bounding boxes around options to visually distinguish the buttons (#2746)
- [x] Make consistent left alignment between existing entries/senses and the "New ___" option
- [ ] Currently one can cancel with the ESC key; add an X button in a corner that invokes the same function as ESC does. (#2746)

From D&LS
"14. H [Select Entry] The “Select an Entry” box was confusing:  
Put a box around the “clickable items” so we can see how many options are on the card (when we click off, there are no boundaries and it just looks like one item).
Make the vernacular have the same indentation as “New Entry”.
Instead of “Select an Entry”, maybe say “Select an option”, since “New Entry” isn’t an entry….
Add a Cancel option (or “none of the above”?)."

imnasnainaec:
Resolves #1342

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2746)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2746)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #2866, opened 2024-01-02
[DataEntry] Auto merge new entry with empty selected dup sense

### LINKED ISSUE #1722: When adding sense to entry with empty sense, allow auto-merging with that sense.
**Is your feature request related to a problem? Please describe.**

Sometimes data is imported into The Combine with entries that ~~only have a vernacular form but have no senses. By default, these entries will each~~ have a single sense with no content (no gloss, nor definition, nor semantic domain).

During data entry (when vernacular autocomplete is turned on for the project), a person might enter a word with the same vernacular form as one of the existing blank-sense entries. When the select-an-entry popup appears, if they select the existing entry, then we end up with whatever gloss they enter as a new sense alongside the empty sense.

**Describe the solution you'd like**

When the only sense available is completely blank, the second popup (select-a-sense) shouldn't even be an option. Rather, the submission of a new gloss should cause the empty sense to be automatically overwritten.

**Acceptance Tests**

Since it is impossible to create a word with an empty sense within The Combine, we need a custom lift file to import for testing this.

imnasnainaec:
Resolves #1722

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2866)


jasonleenaylor (approved):
> Reviewed 1 of 3 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2866)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #2895, opened 2024-01-23
Stop audio/flag/note summary acting like a button

imnasnainaec:
Prevent mouse-over color change when there is no click effect.
Also, use `sx` property to pull colors directly from the theme palette.

Acceptance test:
- Go to Data Cleanup > Review Entries
- Record a pronunciation on one of the entries
- Edit that same entry and add a note and a flag with text
- Go to Data Cleanup and click the top "You've completed:" item (which should be a "Review Entries" tile with "Number of entries edited: 1")
- Visually evaluate the audio/note/flag icons at the top of the result word:
  - They should be align in the top corner and with reasonable padding/margins
  - They should not change color when hovering the cursor
  - The note and flag show have a text tooltip on hover
  - They shouldn't do anything when clicked

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2895)


jasonleenaylor (approved):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/2895)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3067, opened 2024-04-22
[Login] Fix username error bug; Add pw error checks

imnasnainaec:
Fixes bug with the username field:
- click/tab into the username field
- don't type a username or type one that is too short (at most 2 characters)
- click/tab away from the username field
- observe the red error on the field
- return to the username field and add a username at least 3 characters long
- click/tab away from the username field
- observe that the red error persists, even though it shouldn't

Adds on-blur error checking to the password fields.

### DIFF
--- a/src/components/Login/Signup.tsx
+++ b/src/components/Login/Signup.tsx
@@ -102,14 +102,24 @@ export default function Signup(props: SignupProps): ReactElement {
     dispatch(reset());
   }, [dispatch]);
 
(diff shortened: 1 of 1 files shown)


jmgrady (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3067)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3077, opened 2024-04-24
[Recorder] Add backstops for if cursor departs button

### LINKED ISSUE #3073: [AudioRecorder] Recording fails if click/touch ends outside of the button
**Describe the bug**

If you click-and-hold (or, on a touch screen, press-and-hold) on the recording button, it continues to record even if the click/touch exits the button. However, you have to end the click inside the button for it to work. If you release outside the button, it continues to record *AND* if you click on the button to end it, the recording is lost and replaced with a new, truncated recording. The only way to recover the recording if you release outside the button is to click outside the button and move your cursor back to the button before releasing (but this doesn't work with touch).

**Expected behavior**
Two parts:

- The recording should end when the mouse releases even if it's outside the button. ~Perhaps it should end when the mouse leaves at all, or if that's too sensitive, an invisible wider radius could possibly be added.~ This might be approximated with an onBlur.
- When the recording button is clicked, if there's an active recording, that should be finished (rather than / before) a new recording started.

imnasnainaec:
Mostly fixes #3073 

Pro: If the record button is clicked/pressed while a previous recording is still active, that previous recording is finished before a new one is started.
Con: If the record button is clicked/tapped just to end a recording that is still going, a second recording blip will also be created (but at least that is easily deleted).

Pro: If focus shifts away from the button, that will also activate the end of the recording. This works if a user clicks elsewhere.
Con: This doesn't always work if the user taps/presses elsewhere, because the initial recording touch doesn't consistently shift focus to the button in the first place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3077)

### DIFF
--- a/src/components/Pronunciations/AudioRecorder.tsx
+++ b/src/components/Pronunciations/AudioRecorder.tsx
@@ -25,11 +25,19 @@ export default function AudioRecorder(props: RecorderProps): ReactElement {
   const recorder = useContext(RecorderContext);
   const { t } = useTranslation();
 
-  function startRecording(): void {
+  async function startRecording(): Promise<void> {
+    // Prevent starting a recording before a previous one is finished.
+    await stopRecording();
+
     recorder.startRecording();
   }
 
   async function stopRecording(): Promise<void> {
+    // Prevent triggering this function if no recording is active.
+    if (!recorder.isRecording()) {
+      return;
+    }
+
     if (props.onClick) {
       props.onClick();
     }
--- a/src/components/Pronunciations/Recorder.ts
+++ b/src/components/Pronunciations/Recorder.ts
@@ -14,6 +14,10 @@ export default class Recorder {
       .catch((err) => this.onError(err));
   }
 
+  isRecording(): boolean {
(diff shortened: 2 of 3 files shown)


jasonleenaylor (approved):
I think this is a safe step in the right direction, the first Con seems not ideal but it will interact with the improvement @jmgrady was considering of displaying a message and removing accidental recordings. I'm not sure if the interaction will be useful or confusing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3077#-:-NwKuiJP6jJKvxlMKuR-:b-3hipdw)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

jasonleenaylor (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3109, opened 2024-05-02
[SummarySenseCard] Add glosses and bg color

### LINKED ISSUE #3104: [SummarySenseCard] Add glosses, bg color
**Is your feature request related to a problem? Please describe.**

The SenseSummaryCard could better serve the Review Entries dialog for editing a word with the following two additions:

- [x] A concatenation of the sense glosses, as they appear in the Glosses column of the table
- [x] A yellow background if any of the senses are being moved/edited/deleted

imnasnainaec:
Resolves #3104

### DIFF
--- a/src/components/WordCard/SensesTextSummary.tsx
+++ b/src/components/WordCard/SensesTextSummary.tsx
@@ -1,0 +1,75 @@
+import { type ReactElement } from "react";
+
+import { type Sense } from "api/models";
+import { TypographyWithFont } from "utilities/fontComponents";
+
+interface SensesTextSummaryProp {
+  definitionsOrGlosses: "definitions" | "glosses";
+  maxLengthPerSense?: number;
+  senses: Sense[];
+}
+
+export default function SensesTextSummary(
+  props: SensesTextSummaryProp
+): ReactElement {
+  const ellipses = "[...]";
+  const interSenseSep = " | ";
+  const intraSenseSep = "; ";
+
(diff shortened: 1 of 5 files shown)


imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3109)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @jmgrady)

___
*[`src/components/WordCard/index.tsx` line 120 at r3](https://reviewable.io/reviews/sillsdev/TheCombine/3109#-O2_wCtW0VfJirlWOeHO:-O2a8MD7EqgO7Ic4IxJt:b-896fix) ([raw file](https://github.com/sillsdev/TheCombine/blob/3c6aa9d7d7169acac57d70878e10d8fc0c9aa9db/src/components/WordCard/index.tsx#L120)):*

Why is this a property?  It looks to me as though it is always set to `true`.  Is there a use case that would suggest we would want a `SummarySenseCard` without glosses?

Removing the property should simplify the code and the test code so it is worth considering.

If it is not removed, then the test code should be updated to test both cases.
</blockquote></details>

Done.

jmgrady (approved):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3109)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3194, opened 2024-06-28
Add tighter control on project owner management

### LINKED ISSUE #3114: [Project Settings] Edge case: brand new owner can remove themselves
A project administrator (e.g., Simon) can remove themselves from a project.
A project owner (e.g., Garfunkel) can transfer project ownership to a project administrator (e.g., Simon).

If admin Simon opens project settings, Users tab; then on a different computer owner Garfunkel transfers project ownership to Simon; then new owner Simon who still sees himself as just admin Simon removes himself from the project without refreshing the page, he is successful. Result: Garfunkel is a project admin and the project has no owner.

This is a ridiculous edge case, but it shows there could be some safeguard added in the backend to prevent project owner removal without replacement.

imnasnainaec:
Resolves #3114

### DIFF
--- a/Backend/Controllers/UserRoleController.cs
+++ b/Backend/Controllers/UserRoleController.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using BackendFramework.Helper;
 using BackendFramework.Interfaces;
@@ -147,6 +148,13 @@ namespace BackendFramework.Controllers
                 return NotFound(projectId);
             }
 
+            // Prevent a second project owner
+            if (userRole.Role == Role.Owner
+                && (await _userRoleRepo.GetAllUserRoles(projectId)).Any((role) => role.Role == Role.Owner))
+            {
+                return Forbid("This project already has an owner");
+            }
+
             await _userRoleRepo.Create(userRole);
             return Ok(userRole.Id);
         }
@@ -183,6 +191,12 @@ namespace BackendFramework.Controllers
                 return NotFound(userRoleId);
             }
 
+            // Prevent deleting the project owner.
+            if (userRole.Role == Role.Owner)
+            {
+                return Forbid("Cannot use this function to remove the project owner's role");
+            }
+
             // Prevent deleting role of another user who has more permissions than the actor.
             if (!await _permissionService.ContainsProjectRole(HttpContext, userRole.Role, projectId))
             {
@@ -212,6 +226,11 @@ namespace BackendFramework.Controllers
                 return Forbid();
             }
 
+            // Prevent making a new project owner.
+            if (projectRole.Role == Role.Owner)
+            {
+                return Forbid("Cannot use this function to give a user the project owner role");
+            }
             // Prevent upgrading another user to have more permissions than the actor.
             if (!await _permissionService.ContainsProjectRole(HttpContext, projectRole.Role, projectId))
             {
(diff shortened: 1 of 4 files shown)


Backend/Controllers/UserRoleController.cs
```diff
@@ -263,5 +287,95 @@ public async Task<IActionResult> UpdateUserRole(
...
+                newResult = await _userRoleRepo.Update(newRoleId, newUsersRole);
+            }
+            if (newResult != ResultOfUpdate.Updated)
+            {
+                return StatusCode(StatusCodes.Status304NotModified, newRoleId);
+            };
+
+            // Change the old owner to a project admin
+            oldUserRole.Role = Role.Administrator;
+            var oldResult = await _userRoleRepo.Update(oldRoleId, oldUserRole);
```
hahn-kev, line 373: if this fails doesn't that mean there will be 2 owners? is this work being done in some kind of db transaction?

imnasnainaec: @hahn-kev You are correct that this failing could result in two owners. The primary motive for this pr was an edge case that could result in 0 owners. While much of this pr is also avoiding multiple owners, I think it's a less critical problem for that unlikely event to happen.

Though perhaps some sort of aggregate pipeline transaction would be the right way to do this. I'll poke at that idea.

imnasnainaec: One could use Transactions (https://www.mongodb.com/docs/drivers/csharp/current/fundamentals/transactions/) to make multiple changes and commit them all at once, but it appears to me to require a major refactor of how we interface with the database.

imnasnainaec: @hahn-kev Thanks for introducing me to transactions.

hahn-kev: You're welcome. 

Yeah transactions are definitely the way to go when you want all your changes to either fail or succeed together.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3194)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hahn-kev and @imnasnainaec)

___
*[`Backend/Controllers/UserRoleController.cs` line 373 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/3194#gh-1774773492:-O87l3zSDnncwxd1KOTi:b-na2zqe) ([raw file](https://github.com/sillsdev/TheCombine/blob/87256914c1c0007401f43d4d61d0bb4c446a667d/Backend/Controllers/UserRoleController.cs#L373)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

You're welcome. 

Yeah transactions are definitely the way to go when you want all your changes to either fail or succeed together. 
</blockquote></details>

Since transactions are not supported in Standalone MongoDB deployments, we'll hold off implementing them until some other month after we've switched to Replica Set deployment.

jasonleenaylor (approved):
:lgtm: The transactions will wait. This is a good step in the meantime.

> Reviewed 4 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3194#-:-O8bhA6h2gF36_eTMFhz:b-qg2vxe)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

hahn-kev (approved):

PR #3232, opened 2024-07-05
Update TS target to ES6 (which we already use)

imnasnainaec:
We're already using ES6 features (e.g., arrow functions and `let`/`const`), so this change just helps the IDE know what Typescript syntax is allowable.

Before this change, my IDE was complaining about the use of the `/u` RegEx ending in `src/utilities/spellChecker.ts`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3232)


jmgrady (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3232)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3290, opened 2024-08-01
Clear duplicate selection when user returns to vernacular input

imnasnainaec:
Fixes #3259 (both gives the expected behavior and fixes the odd bug)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3290)

PR #3296, opened 2024-08-06
Revert #3290: Clear vern selection when user returns to vern field

imnasnainaec:
This reverts commit 474ec40e429bdc14d0b68448b2766ae624eed7dd.

That bugfix created a different bug, where the vern dialog duplicate selection sometimes returns focus to the vern field, which clears what was just selected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3296)


imnasnainaec:
Rather than retract, I think I've fixed the issue in #3297.

PR #3360, opened 2024-09-17
Standalone installer improvements

jmgrady:
These changes improve the standalone installer as follows:

- This change adds a `server` option so that installer can be used to setup a NUC (or other device) running as a server.  If the `server` option is specified, then the Combine is configured to start at boot time.
- The Python virtual environment is removed from the installer.  This means that more code must be downloaded during the installation but the installer can run on a system that has a version of Python3 installed that is different from the system that was used to create the installer.
- The script to make the installer will now generate the `Chart.yaml` files from their templates for all of the `helm` charts before building the packages.  Before this change, the installer could only be created from a working directory that had been used to install _The Combine_ previously.

### DIFF
--- a/deploy/scripts/combine_charts.py
+++ b/deploy/scripts/combine_charts.py
@@ -70,6 +70,6 @@ def generate(version: str, aws_login_version: str = default_aws_login_version) -
 
 
 if __name__ == "__main__":
-    """Allow calling from the command line for testing, etc."""
+    """Allow calling from the command line."""
     args = parse_args()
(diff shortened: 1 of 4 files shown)


imnasnainaec (approved):
> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3360)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #3369, opened 2024-09-27
Expose mongo transactions

hahn-kev:
From [this](https://github.com/sillsdev/TheCombine/pull/3194#discussion_r1774773492) discussion and a follow up with Jason he mentioned that transactions haven't been exposed to utilize them in mongo.

In order for transactions to work across repos there needs to only be one Mongo client in a given request, I've refactored all the type specific contexts to get the IMongoDb from the new IMongoDbContext which is a [scoped service](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#scoped)  so there's only one instance per request.

Then I changed a couple methods in the Merge service to make changes inside transactions. You'll notice that the transaction is not aborted in our user code, this is because the transaction is disposable and if it's disposed before it's committed then it will be aborted, so if an exception is thrown between Begin and Commit then the changes made will be rolled back.

Feel free to suggest changes, I don't really love how I had to mock out and expose the transactions via the IMongoDbContext which also contains the IMongoDatabase, but I didn't really feel like making a separate service just for transactions, @jasonleenaylor let me know what you think there if we want to split those out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3369)

### DIFF
--- a/Backend.Tests/Controllers/MergeControllerTests.cs
+++ b/Backend.Tests/Controllers/MergeControllerTests.cs
@@ -14,6 +14,7 @@ namespace Backend.Tests.Controllers
         private IMergeBlacklistRepository _mergeBlacklistRepo = null!;
         private IMergeGraylistRepository _mergeGraylistRepo = null!;
         private IWordRepository _wordRepo = null!;
+        private IMongoDbContext _mongoDbContext = null!;
         private IMergeService _mergeService = null!;
         private IPermissionService _permissionService = null!;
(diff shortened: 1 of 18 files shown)


imnasnainaec:
I extended this pr's branch with some cleanup and a fix (adding the missing `StartTransaction`): https://github.com/sillsdev/TheCombine/tree/mongo-context

imnasnainaec:
Replaced by #3484

PR #3370, opened 2024-09-27
Fix shebang for Python scripts

jmgrady:
Change the "shebang" for Python scripts from:
```
#!/usr/bin/env python3
```
to
```
#!/usr/bin/env python
```

The original shebang was set so that Linux users can run the Python scripts by just providing the script name, e.g. `./deploy/scripts/app_release.py --get`.

Both of these shebang lines work because the `venv` module sets up `python3` and `python` to point to the version of Python in the virtual environment.  However, when the `venv` is created on Windows, only the `python` alias is created.  This means that the scripts fail when run from a `Git Bash`  session.

The new shebang works for both environments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3370)

### DIFF
--- a/deploy/scripts/app_release.py
+++ b/deploy/scripts/app_release.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 from __future__ import annotations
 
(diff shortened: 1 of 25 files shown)


imnasnainaec (approved):
> Reviewed 25 of 25 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3370)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jmgrady)

PR #3443, opened 2024-11-12
[build] Fix image load bug

### LINKED ISSUE #3439: setup_combine.py fails when run from a detached HEAD
If you are on a detached HEAD rather than a named git branch, running `setup_combine.py` will result in `public/scripts/release.js` with an invalid version tag (e.g., `v2.2.0-.68`). There needs to be a non-empty string after the `-`.

### LINKED ISSUE #3442: [build] Docker failing to load images in dev build
**Describe the bug**

When running `deploy/scripts/build.py`, the builds are not automatically loaded in Docker as before.

**To Reproduce**

0. Delete all local Docker images, containers, builds for The Combine
1. Start a Python virtual environment `(venv)`
2. Run `python deploy/scripts/setup_cluster.py --type development`
3. Run `python deploy/scripts/build.py`
4. Run `docker images` and see that the desired images aren't present
3. Run `kubectl -n thecombine get deployments` and see `0/1` in the READY column (should be `1/1`)
4. Run `kubectl -n thecombine get pods` to get failed pod names (e.g. `backend-9d66df89b-m25pq`)
5. Run (e.g.) `kubectl -n thecombine logs backend-9d66df89b-m25pq` to see that the image failed to pull
6. Run `python deploy/scripts/build.py --debug` and you may notice warning like 
> WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

It seems some Docker or buildx update no longer loads our builds by default and we need to add `--load` to the `docker buildx build [...]` commands.

imnasnainaec:
Fixes #3442

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3443)

### DIFF
--- a/deploy/scripts/app_release.py
+++ b/deploy/scripts/app_release.py
@@ -31,7 +31,7 @@ def get_release() -> str:
                 num_commits = match[2]
             # Get the branch name
             result = run_cmd(["git", "branch", "--show-current"], chomp=True)
-            branch_name = re.sub("[/_]+", "-", result.stdout)
+            branch_name = re.sub("[/_]+", "-", result.stdout) or "HEADLESS"
             return f"{release_string}-{branch_name}.{num_commits}"
     message = f"Unrecognized release value in tag: {result.stdout}"
     raise ValueError(message)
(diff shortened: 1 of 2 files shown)


imnasnainaec:
Issue was closed.

PR #3523, opened 2025-01-15
[MergeDups] Add failsafe for duplicate guids and other bad merge sets

### LINKED ISSUE #3471: Problem with same sense guid in multiple entries
The same sense guid should not appear in different senses. There's a tool that runs with FLEx send/receive to prevent that. However, The Combine doesn't prevent it.

Moreover, this creates an issue with Merge Duplicates. If you have two entries with the same vernacular form and both with a sense that has the same sense guid and if you try to merge one of those senses into the other then click "Save & Continue", the merge won't happen. (Work-around: delete one of the senses instead of trying to merge them.)

- Can we mitigate having the same sense guid in multiple places?
- When it happens, what fail-safe do we need in the merge tool for it to work as expected?

imnasnainaec:
Resolves #3471

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3523)

### DIFF
--- a/src/components/DataEntry/DataEntryTable/index.tsx
+++ b/src/components/DataEntry/DataEntryTable/index.tsx
@@ -1005,11 +1005,12 @@ export default function DataEntryTable(
       const oldWord = state.recentWords[index].word;
       text = text.trim();
       if (text !== oldWord.note.text) {
-        const note: Note = { ...oldWord.note, text };
+        const language = oldWord.note.language || analysisLang.bcp47;
+        const note: Note = { ...oldWord.note, language, text };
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3523#-:-OGfmLgZ-9RNeZyU5mpj:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3524, opened 2025-01-16
update setup dotnet

hahn-kev:
was v4.0.1 to v4.2.0, this should fix the workflow failures. This was caused by the [this](https://devblogs.microsoft.com/dotnet/critical-dotnet-install-links-are-changing/)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3524)

PR #3527, opened 2025-01-16
Update setup-dotnet to v4.2.0 and clear old endpoints

imnasnainaec:
Supersedes #3524

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3527)


hahn-kev (commented):
looks good to me, it does look like `ci.dot.net` was added but I suspect that would only be used if we were to try and use a nightly build in setup-dotnet so it's probably not worth adding

> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3527#-:-OGm73Xf1f8kWelWj3bc:bq2kczw)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3626, opened 2025-02-27
Update import/export handling of Vernacular

### LINKED ISSUE #3532: [FLEx] Citation form overwritten by lexeme form in round-trip
The Combine's import prefers citation form over lexeme form if the former is present in the LIFT file. However, the `vernacular` in The Combine doesn't distinguish between citation form and lexeme form, and exports as lexeme form. So if a FLEx entry has different citation and lexeme forms, a round-trip through The Combine will cause the entry's lexeme form in FLEx to be overwritten with its citation form.

A possible fix:

1. Add a Boolean field to the `Word` model for `usingCitationForm`
2. Only set to `true`: (a) on import; and (b) if the citation form is used
3. Change to `false` if the vernacular is ever changed in The Combine
4. If it's `true` on export, export vernacular to citation form instead of lexeme form

imnasnainaec:
Fixes #3532

- Track if citation form instead of lexical form was used for `Vernacular` (and export to citation form if imported with citation form and `Vernacular` is unchanged)
- When a LIFT file is selected in project settings for import into an existing project, compare the LIFT's ldml languages against the projects vernacular language (and require user confirmation if the languages don't match)
- Only import words that have a citation or lexical form in the same language as the projects vernacular language (and toast the number of words imported)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3626)


jasonleenaylor:
*[`Backend/Services/LiftService.cs` line 775 at r1](https://reviewable.io/reviews/sillsdev/TheCombine/3626#-OKTNO8MCVPQ_24C7wbS:-OKTNO8MCVPQ_24C7wbT:bwmz4ei) ([raw file](https://github.com/sillsdev/TheCombine/blob/399064aa9a74da7c3355ceba190f3e5f1e81751e/Backend/Services/LiftService.cs#L775)):*
> ```Smalltalk
>                 if (string.IsNullOrWhiteSpace(vern))
>                 {
>                     vern = entry.CitationForm.FirstOrDefault(x => !string.IsNullOrWhiteSpace(x.Value.Text)).Value?.Text;
> ```

What if we didn't do this?

Backend.Tests/Models/WordTests.cs
```diff
@@ -47,9 +47,10 @@ public void TestNotEqualsNote()
         [Test]
         public void TestHashCode()
         {
-            Assert.That(
-                new Word { Guid = _commonGuid, Vernacular = "1" }.GetHashCode(),
-                Is.Not.EqualTo(new Word { Guid = _commonGuid, Vernacular = "2" }.GetHashCode()));
+            var hashCode = new Word { Guid = _commonGuid, Vernacular = "1" }.GetHashCode();
+            Assert.That(hashCode, Is.Not.EqualTo(new Word { Guid = _commonGuid, Vernacular = "2" }.GetHashCode()));
+            Assert.That(hashCode, Is.Not.EqualTo(
+                new Word { Guid = _commonGuid, Vernacular = "1", UsingCitationForm = true }.GetHashCode()));
```
hahn-kev, line 53: you've got a bug in your test. `UsingCitationForm` is not referenced in `GetHashCode` for `Word`, which is a bug. But the **real** issue is that this test didn't catch that. Here's a failing test that should pass which demonstrates the issue
```C#
var word1 = new Word { Guid = _commonGuid, Vernacular = "1" };
var word2 = new Word { Guid = _commonGuid, Vernacular = "1" };
var word1HashCode = word1.GetHashCode();
var word2HashCode = word2.GetHashCode();
Assert.That(word1HashCode, Is.EqualTo(word2HashCode));
```
the hash codes aren't the same even though they should be. I think this is because the `HashCode` class just calls `GetHashCode` on the objects passed in, and most objects (`List<T>` for example) don't overload GetHashCode, which means that the hash code is unique per instance of the list, not based on it's contents. Eg: `new List<string>().GetHashCode() != new List<string>().GetHashCode()` this is expected and by design as a hash code is expected to not change for a given object over it's lifetime, otherwise it wouldn't work as a key in a dictionary for example. I could go more into this but I'll stop before I write an essay.

imnasnainaec: https://github.com/sillsdev/TheCombine/pull/3660

jasonleenaylor (approved):
> Reviewed 8 of 8 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3626)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

imnasnainaec (commented):
> Dismissed @hahn-kev from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3626)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3660, opened 2025-03-05
Don't override Equals, GetHashCode

imnasnainaec:
We thought they were only used in testing, but they snuck into production via 4 uses of `.Remove` as well as uses of `.Contains` in other model methods.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3660)


jasonleenaylor (commented):
I had two comments worth considering, but overall :lgtm:

> Reviewed 33 of 47 files at r1, 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3660#-:-OL0zb53-_qytVQaxohz:bdwq7k2)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

___
*[`Backend/Services/WordService.cs` line 109 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/3660#-OKgykB15MRnm2YmAvaX:-OKgykB15MRnm2YmAvaY:bwf30tl) ([raw file](https://github.com/sillsdev/TheCombine/blob/e3a9d6152fe7317cc3a868d06da400c209772f14/Backend/Services/WordService.cs#L109)):*
> ```Smalltalk
> 
>             var rmCount = wordWithAudioToDelete.Audio.RemoveAll(a => a.FileName == fileName);
>             if (rmCount == 0)
> ```

There is nothing at all wrong with this change, but I found it less clear than the original. So unless we expect Audio to have duplicate entries with the same filename I would slightly prefer the original. Should be bring @andracc in as a tie breaker? :smile: 



_Code quote:_
```Smalltalk
            var rmCount = wordWithAudioToDelete.Audio.RemoveAll(a => a.FileName == fileName);
            if (rmCount == 0)
```
___
*[`Backend.Tests/Controllers/UserRoleControllerTests.cs` line 241 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/3660#-OKh-I5x4KVAOhQKBEj0:-OKh-I5yACjD3xr-TgNU:b-c89bdy) ([raw file](https://github.com/sillsdev/TheCombine/blob/6cb624c93051c40d3e8abb22123bb637392b4837/Backend.Tests/Controllers/UserRoleControllerTests.cs#L241)):*
> ```Smalltalk
> 
>         [Test]
>         public async Task TestUpdateUserRoleNoChange()
> ```

So this test no longer had any value?

Backend.Tests/Models/MergeWordsTests.cs
```diff
@@ -1,72 +1,15 @@
...
-
         [Test]
-        public void TestHashCode()
+        public void TestClone()
         {
-            Assert.That(
-                new MergeUndoIds { ParentIds = _parentIds1, ChildIds = _childIds1 }.GetHashCode(),
-                Is.EqualTo(new MergeUndoIds { ParentIds = _parentIds1, ChildIds = _childIds1 }.GetHashCode()));
+            var merge = new MergeUndoIds { ParentIds = ["parent1", "parent2"], ChildIds = ["child1", "child2"] };
+            Assert.That(merge.Clone(), Is.EqualTo(merge).UsingPropertiesComparer());
```
hahn-kev, line 12: this test will pass even if the clone is not a deep clone. So you might want to check that Parent and Child Ids don't refer to the same lists. Just a thought though.

hahn-kev (approved):
I'll bet that felt good to delete all that code

Backend.Tests/Models/MergeWordsTests.cs
imnasnainaec: Ah, yes. This is why you use
```
            assertionChain
                .ForCondition(!ReferenceEquals(comparands.Subject, comparands.Expectation))
                .BecauseOf(context.Reason)
                .FailWith("Subject and Expectation for {0} should not reference the same instance in memory.", context.CurrentNode.Description);
```
I think for now I'll leave it and make an issue for considering that with all our Clone tests. The main purpose of the tests is to make sure no property was left out, but it would be valuable to make sure they've all been truly cloned.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3660)* status: 44 of 47 files reviewed, 1 unresolved discussion (waiting on @hahn-kev and @jasonleenaylor)

___
*[`Backend.Tests/Controllers/UserRoleControllerTests.cs` line 241 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/3660#-OKh-I5x4KVAOhQKBEj0:-OL4mxUCExhOPE0e6ZG3:b5fe3ys) ([raw file](https://github.com/sillsdev/TheCombine/blob/6cb624c93051c40d3e8abb22123bb637392b4837/Backend.Tests/Controllers/UserRoleControllerTests.cs#L241)):*

So this test no longer had any value?
</blockquote></details>

Very little. Most of our Repository Mocks don't have a "NotModified" case, and this test depends on the
```
            if (foundUserRole.ContentEquals(userRole))
            {
                return Task.FromResult(ResultOfUpdate.NoChange);
            }
```
case of the `Update` method in `UserRoleRepositoryMock`, which was the only reason we had `UserRole.ContentEquals` defined. So I thought I'd remove it to have `UserRoleRepositoryMock` more consistent with the other Mocks.
___
*[`Backend/Services/WordService.cs` line 109 at r2](https://reviewable.io/reviews/sillsdev/TheCombine/3660#-OKgykB15MRnm2YmAvaX:-OL44jiF5PDSr8416fSj:bp3aow6) ([raw file](https://github.com/sillsdev/TheCombine/blob/e3a9d6152fe7317cc3a868d06da400c209772f14/Backend/Services/WordService.cs#L109)):*

There is nothing at all wrong with this change, but I found it less clear than the original. So unless we expect Audio to have duplicate entries with the same filename I would slightly prefer the original. Should be bring @andracc in as a tie breaker? :smile:
</blockquote></details>

Reverted.

Backend.Tests/Models/MergeWordsTests.cs
imnasnainaec: https://github.com/sillsdev/TheCombine/issues/3669

jasonleenaylor (approved):
:lgtm:

> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3660#-:-OL5C5-78cLlEZ0hXJlQ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

imnasnainaec (commented):
> Dismissed @hahn-kev from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3660)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

PR #3670, opened 2025-03-11
Export/import flag as custom field

### LINKED ISSUE #3264: [Export] Include flags
Requested by Hasso:

> Data Loss: red flags are not exported, but I have seen no reason not to expect them to be exported. This is important because:
>    - I expect them to be available when working in FLEx or WeSay
>    - I may wish to round trip with FLEx, as finding fuzzy duplicates is easier in The Combine, but The Combine cannot clean up entries with data in custom fields (we had several duplicates like this from a Paratext import)
>    - We needed to switch servers twice due to equipment availability, and would have moved our data again to the cloud for post-workshop cleanup if we didn’t have custom fields

imnasnainaec:
Resolves #3264

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3670)

### DIFF
--- a/Backend.Tests/Controllers/LiftControllerTests.cs
+++ b/Backend.Tests/Controllers/LiftControllerTests.cs
@@ -69,7 +69,7 @@ namespace Backend.Tests.Controllers
                 _wordRepo, _projRepo, _permissionService, _liftService, _notifyService, _logger);
 
             _logger = new MockLogger();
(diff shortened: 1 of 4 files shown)


imnasnainaec:
We need https://github.com/sillsdev/libpalaso/issues/1413 first.

PR #3718, opened 2025-04-08
Normalize project data text (NFC)

### LINKED ISSUE #3713: Unicode not normalized
A project has two words with identical vernacular form, "bôh" and "bôh", however, the first is (U+0062 U+006F U+0302 U+0068) and the second is (U+0062 U+00F4 U+0068). We have not normalized our strings so the U+006F U+0302 ("o" followed by a combining circumflex) is different from U+00F4 (ô) and those should be the same.

We do normalize on export (NFC or NFKC... I'll need to confirm which), but we need to normalize on data input within the combine, so that the Merge Duplicates tool doesn't miss these. We also need to normalize the column filter in Review Entries.

imnasnainaec:
Resolves #3713

### DIFF
--- a/Backend/Services/LiftService.cs
+++ b/Backend/Services/LiftService.cs
@@ -19,6 +19,7 @@ using SIL.Lift.Parsing;
 using SIL.Text;
 using SIL.WritingSystems;
 using Xabe.FFmpeg;
+using static System.Text.NormalizationForm;
 using static SIL.DictionaryServices.Lift.LiftWriter;
 
 namespace BackendFramework.Services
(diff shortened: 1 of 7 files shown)


jasonleenaylor (approved):
:lgtm:

> Reviewed 2 of 9 files at r1, 24 of 24 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3718#-:-ONpbT630nbSPeeQLNKi:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @imnasnainaec)

Issue #3720, opened 2025-04-10
MergeDuplicate errors (524) if duplicate-finding takes longer than 100 seconds


imnasnainaec:
**Describe the bug**

The duplicate finder is quick when finding sets of words with identical vernacular form. It takes a lot longer when those are exhausted, because next are sets of similar-vernacular-form words, which need to have their similarities ordered. This can take several minutes on a multi-thousand word project. If it takes over 100 seconds, we get a 524 error, which is Cloudflare's custom timeout error. We need to do one of the following:

- Rewrite the merge tool
  - Signal when duplicates are found (as we do with the export)?
  - Poll until it's done?
- ~Update our Cloudflare settings with this particular function to not timeout (for 10 minutes?)~ [only available for Enterprise plans]

PR #3726, opened 2025-04-17
[MergeDups] Early-return if full dups list below similarity threshold

imnasnainaec:
Ameliorate  #3720

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3726)


imnasnainaec:
This give some results quickly, but doesn't fix how long it takes to find similar words that are all late in the list of words.

PR #3756, opened 2025-04-28
For non-empty project check, just query database for 1 word

### LINKED ISSUE #3751: Improve database query for checking if project is empty
There are two places where, to check if a project is empty, we ask the database to return the full set of words or of frontier words for a project and then have the backend verify the returned list is nonempty. Surely we can just query the database for one result with the right project id.

- https://github.com/sillsdev/TheCombine/blob/v2.3.1/Backend/Repositories/WordRepository.cs#L164-L171
- https://github.com/sillsdev/TheCombine/blob/v2.3.1/Backend/Controllers/LiftController.cs#L331-L337

imnasnainaec:
Resolves #3751 

Reduces delay before the Export button in enabled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3756)

### DIFF
--- a/Backend/Interfaces/IWordRepository.cs
+++ b/Backend/Interfaces/IWordRepository.cs
@@ -12,8 +12,8 @@ namespace BackendFramework.Interfaces
         Task<List<Word>> Create(List<Word> words);
         Task<Word> Add(Word word);
         Task<bool> DeleteAllWords(string projectId);
-        Task<bool> IsWordsNonempty(string projectId);
-        Task<bool> IsFrontierNonempty(string projectId);
+        Task<bool> HasWords(string projectId);
+        Task<bool> HasFrontierWords(string projectId);
         Task<bool> IsInFrontier(string projectId, string wordId);
         Task<List<Word>> GetFrontier(string projectId);
         Task<List<Word>> GetFrontierWithVernacular(string projectId, string vernacular);
--- a/Backend/Interfaces/IWordRepository.cs
+++ b/Backend/Interfaces/IWordRepository.cs
@@ -12,6 +12,7 @@ namespace BackendFramework.Interfaces
         Task<List<Word>> Create(List<Word> words);
         Task<Word> Add(Word word);
         Task<bool> DeleteAllWords(string projectId);
+        Task<bool> IsWordsNonempty(string projectId);
         Task<bool> IsFrontierNonempty(string projectId);
         Task<bool> IsInFrontier(string projectId, string wordId);
         Task<List<Word>> GetFrontier(string projectId);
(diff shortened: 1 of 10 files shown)


Backend/Repositories/WordRepository.cs
```diff
@@ -161,13 +164,23 @@ public async Task<Word> Add(Word word)
             return word;
         }
 
+        /// <summary> Checks if Words collection is nonempty for specified <see cref="Project"/> </summary>
+        public async Task<bool> IsWordsNonempty(string projectId)
+        {
+            using var activity =
+                OtelService.StartActivityWithTag(otelTagName, "checking if WordsCollection is nonempty");
+
+            return await _wordDatabase.Words.Find(GetAllProjectWordsFilter(projectId)).Limit(1)
```
hahn-kev, line 173: might I suggest [`AnyAsync`](https://mongodb.github.io/mongo-csharp-driver/3.3.0/api/MongoDB.Driver/MongoDB.Driver.Linq.MongoQueryable.AnyAsync.html)?

Backend/Interfaces/IWordRepository.cs
```diff
@@ -12,6 +12,7 @@ public interface IWordRepository
         Task<List<Word>> Create(List<Word> words);
         Task<Word> Add(Word word);
         Task<bool> DeleteAllWords(string projectId);
+        Task<bool> IsWordsNonempty(string projectId);
```
hahn-kev, line 15: just a nit pick on the name. Having names in the negative can be tricky to reason about in boolean logic. Additionally the only place you're using this right now is in the negative case, which makes a double negative, so `!NonEmpty`. it might be more straightforward to change it to `IsWordsEmpty` then it's `IsEmpty` or `!IsEmpty`. If you want to keep the current meaning, I would call it `HasWords` or `AnyWords` then you don't have a double negative when you want to see if it's empty.
Feel free to ignore this unsolicited input.

imnasnainaec: Done.

Backend/Repositories/WordRepository.cs
imnasnainaec: Done.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3756)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @andracc and @hahn-kev)

imnasnainaec (commented):
> Dismissed @hahn-kev from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3756)* status: 6 of 10 files reviewed, all discussions resolved (waiting on @andracc and @imnasnainaec)

PR #3793, opened 2025-05-13
[Data Cleanup] Speed up check for deferred duplicates

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3793)


Backend/Controllers/MergeController.cs
```diff
@@ -152,6 +152,21 @@ public IActionResult RetrievePotentialDuplicates()
             return dups is null ? BadRequest() : Ok(dups);
         }
 
+        /// <summary> Get whether user has graylist entries. </summary>
+        /// <param name="projectId"> Id of project in which to search the frontier for potential duplicates. </param>
+        /// <param name="userId"> Id of user whose merge graylist is to be used. </param>
+        [HttpGet("hasgraylist/{userId}", Name = "HasGraylistEntries")]
+        [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(bool))]
+        public async Task<IActionResult> getGraylistEntries(string projectId, string userId)
```
hahn-kev, line 160: the name `getGraylistEntries` should be capitalized

Backend/Repositories/WordRepository.cs
```diff
@@ -189,6 +189,19 @@ public async Task<bool> IsInFrontier(string projectId, string wordId)
...
+        /// <summary> Checks if given words are in the project Frontier. </summary>
+        /// <param name="projectId"> Id of project to check in. </param>
+        /// <param name="wordIds"> Ids of words to check for. </param>
+        /// <param name="count"> Minimum number of words required. </param>
+        public async Task<bool> AreInFrontier(string projectId, List<string> wordIds, int count)
+        {
+            using var activity = OtelService.StartActivityWithTag(otelTagName, "checking if Frontier contains words");
+
+            var words = await _wordDatabase.Frontier.Find(GetProjectWordsFilter(projectId, wordIds))
+                .Limit(count).ToListAsync();
```
hahn-kev, line 201: might I suggest `CountAsync`?

Backend/Controllers/MergeController.cs
imnasnainaec: Done.

Backend/Repositories/WordRepository.cs
imnasnainaec: You might... it's a good suggestion. I opted for `Find` so I could use `.Limit`, which doesn't work with `CountDocuments`. But at your suggestion, I took a look at https://www.mongodb.com/docs/manual/reference/method/db.collection.countDocuments/ and discovered that it supports the limit in a different way!

imnasnainaec (commented):
> Dismissed @hahn-kev from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3793)* status: 3 of 13 files reviewed, all discussions resolved (waiting on @imnasnainaec)

___
*[`Backend.Tests/Mocks/WordRepositoryMock.cs` line 81 at r3](https://reviewable.io/reviews/sillsdev/TheCombine/3793#-OQJF8E1Ct9vTMCS5G-3:-OQJIBvE6qClK7p6zMf6:b6xuxqv) ([raw file](https://github.com/sillsdev/TheCombine/blob/6faa54684c031eafeed655c65ee9a84133d23059/Backend.Tests/Mocks/WordRepositoryMock.cs#L81)):*
<details><summary><i>Previously, andracc (Andra) wrote…</i></summary><blockquote>

Since `count` represents a minimum and we are not limiting the results here, it looks like we can use `>=` instead of `==`.
</blockquote></details>

We shouldn't have multiple words in the frontier with the same id, though we're not enforcing that in the test setting, so we could go either `>=` or `==` here. I'm inclined to keep it as `==` to closer mimic the expectations of the real thing.

PR #3808, opened 2025-05-20
[DataEntry, TreeView] WIP: Improve narrow window UX

### LINKED ISSUE #3812: [TreeView] Design tree navigation for mobile users
We chose to remove the tree for `xs`-width (<600 px), since those doing RWC will know the domain id and can just type that. However
- Users doing other forms of word collection should still be able to navigate the tree
- On `sm`-width (<900 px), domains with 9 subdomains can still overflow the screen

We need a design that works on narrow windows (300 - 900 px).

Idea from discussion with James Post: Put children in a grid with 3 or 4 columns.

Also consider: breadcrumbs at the top.

### LINKED ISSUE #3813: [DataEntry] Redesign for mobile view
On narrow windows (300-600 px), our Vernacular and Gloss fields are too short, the note/audio/trash buttons are too crowded, and the audio box awkwardly wraps with only one or two recordings.

Discussed some design ideas with James Post. We probably need a condensed view for the Recent Entries and a multi-row mode for adding a New Entry or editing a Recent Entry. This mode could either expand or be in a dialog (like the `EditDialog` and `EditSenseDialog` in Review Entries).

The handling of the Existing Entries (those in the domain from previous sessions) could also be improved.

imnasnainaec:
Based on #3787 (see diff: https://github.com/sillsdev/TheCombine/compare/grid2...tree-view-sm)

Resolves #3812 
Resolves #3813 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3808)


imnasnainaec:
Split into #3816 & #3847

PR #3837, opened 2025-06-03
[SemanticDomainController] Fix GetAllSemanticDomainTreeNodes return type

imnasnainaec:
This function isn't currently used in the frontend.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3837)

### DIFF
--- a/src/api/api/semantic-domain-api.ts
+++ b/src/api/api/semantic-domain-api.ts
@@ -337,7 +337,7 @@ export const SemanticDomainApiFp = function (configuration?: Configuration) {
       (
         axios?: AxiosInstance,
         basePath?: string
-      ) => AxiosPromise<SemanticDomainTreeNode>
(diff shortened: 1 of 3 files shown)


src/api/api/semantic-domain-api.ts
```diff
@@ -478,7 +478,7 @@ export const SemanticDomainApiFactory = function (
     getAllSemanticDomainTreeNodes(
       lang?: string,
       options?: any
-    ): AxiosPromise<SemanticDomainTreeNode> {
+    ): AxiosPromise<Array<SemanticDomainTreeNode>> {
```
hahn-kev, line 481: not sure what AxiosPromise looks like, but since `GetAllSemanticDomainTreeNodes` can return null shouldn't that be accounted for in the return type?

hahn-kev (approved):

PR #3838, opened 2025-06-03
[PermissionService] Remove IsCurrentUserAuthorized redundancies

imnasnainaec:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3838)

### DIFF
--- a/Backend.Tests/Controllers/AvatarControllerTests.cs
+++ b/Backend.Tests/Controllers/AvatarControllerTests.cs
@@ -32,6 +32,8 @@ namespace Backend.Tests.Controllers
         }
 
         private User _jwtAuthenticatedUser = null!;
+        private const string FileName = "combine.png";  // File in Backend.Tests/Assets/
+        private readonly string _filePath = Path.Combine(Util.AssetsDir, FileName);
 
         [SetUp]
         public void Setup()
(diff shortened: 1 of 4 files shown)


Backend/Services/PermissionService.cs
```diff
@@ -41,22 +41,20 @@ private static SecurityToken GetJwt(HttpContext request)
         /// <summary> Checks whether the given user is authorized. </summary>
         public bool IsUserIdAuthorized(HttpContext request, string userId)
         {
-            var currentUserId = GetUserId(request);
-            return userId == currentUserId;
+            return userId == GetUserId(request);
```
hahn-kev, line 44: As this is right now if both of these strings were empty this would return true. Not sure why that would be the case but it would probably be a bug or error and this should return false in that case.

hahn-kev (approved):
the changes look good to me, I'm concerned with some of the ways security is being handled right now but that's not in scope of this PR as you're not really changing anything.

imnasnainaec:
> the changes look good to me, I'm concerned with some of the ways security is being handled right now but that's not in scope of this PR as you're not really changing anything.

@hahn-kev Do tell me more. I'd be glad to improve the security in another pr.

Backend/Services/PermissionService.cs
imnasnainaec: Done.

imnasnainaec (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3838)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @hahn-kev and @imnasnainaec)

PR #3847, opened 2025-06-05
[DataEntryTable] Redesign UX for narrow window

### LINKED ISSUE #3493: [DataEntry] Data entry difficult on mobile
Some feedback from a user having difficulty with data entry:

> The space for entering information is way too small and cluttered, in portrait and (worse) in landscape position of the screen alike.
> There is no “enter” key, so one does not know how to finalize an entry and proceed to a next new entry.

We may want to design a simple (dialog?) component for entering a single word that is designed for small screens, e.g., a minimal version of the Review Entries component for editing an entry.

### LINKED ISSUE #3813: [DataEntry] Redesign for mobile view

imnasnainaec:
Resolves #3493 
Resolves #3813

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/3847)

### DIFF
--- a/src/components/DataEntry/DataEntryTable/RecentEntryCold.tsx
+++ b/src/components/DataEntry/DataEntryTable/RecentEntryCold.tsx
@@ -1,0 +1,74 @@
+import { Edit } from "@mui/icons-material";
+import { Grid2 } from "@mui/material";
+import { ReactElement, memo } from "react";
+
+import { Word, WritingSystem } from "api/models";
+import { NoteButton } from "components/Buttons";
+import { AudioSummary } from "components/WordCard";
+import theme from "types/theme";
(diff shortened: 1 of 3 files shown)


imnasnainaec:
Replaced by #4373

PR #4045, opened 2025-11-26
Use ignore-scripts to guard against supply-chain attacks

imnasnainaec:
This is to reduce our vulnerability to supply-chain attacks.
- Add `--ignore-scripts` to every execution of `npm ci`, `npm i` and `npx`.
- Redundantly create `.npmrc` with `ignore-scripts=true`.
- Redundantly add `npm_config_ignore_scripts: true` in `env:` to every `npm` workflow call.

The redundancies would make it more obvious if a bad pr tries to bypass this.

Also switch our frontend license generator to `npx`, since we don't otherwise need that package and its dependencies.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4045)

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -19,6 +19,8 @@ permissions: # added using https://github.com/step-security/secure-workflows
 jobs:
   lint_build:
     runs-on: ubuntu-latest
+    env:
+      npm_config_ignore_scripts: true
     strategy:
       matrix:
         node-version: [22]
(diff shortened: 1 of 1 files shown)


.github/workflows/frontend.yml
```diff
@@ -41,10 +41,22 @@ jobs:
...
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm ci
-      - run: npm run fmt-frontend-check
-      - run: npm run lint
-      - run: npm run build
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+        env:
+          npm_config_ignore_scripts: true
```
myieye: You could consider putting this at the job level rather than the step level.

PR #4155, opened 2026-02-16
[MergeService] Refactor methods

imnasnainaec:
Also remove no-longer-needed `DeleteFrontierWords` from `WordRepository`.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4155)

## Summary by CodeRabbit

* **Refactor**
  * Merge workflow rewritten for more reliable parent/child handling, deduplication, and batch operations.
  * Undo-merge process validated and hardened to avoid accidental restores.
* **Bug Fixes**
  * Improved handling of ID alignment and history/citation propagation between items.
* **Chore**
  * Removal of bulk frontier-delete API and corresponding test mock (affects bulk-delete behavior).

PR #4178, opened 2026-02-24
[export] Only export as deleted entries explicitly so

imnasnainaec:
If a word is removed from the frontier by some unforeseen bug, we don't want that word deleted from the user's FieldWorks project. So this pr will add a guard against that.

Required database update to go with this pr:

- [ ] Add Deleted copies of any words with GUID missing from the Frontier.

Reason: Prior to #4155, merge children were not marked as deleted in the database, so without the database update, this change would result in the children of merges not being deleted in FW, which would undermine the cleanup done in TC.

Side-effect: We cannot distinguish between words that were wrongfully lost from the fronter and words that were deleted (not merged) with the Merge Duplicates tool. So any of the former (which are so far only known to exist in one project), will be cemented as Deleted.

The database update will be easier after #4179 (and it's associated database update) is complete.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4178)

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected word exports so only words explicitly marked as deleted and no longer active are reported as deleted.
  - Prevented words that are no longer in the Frontier from being incorrectly reported as deleted.

### DIFF
--- a/Backend/Services/LiftService.cs
+++ b/Backend/Services/LiftService.cs
@@ -283,6 +283,8 @@ namespace BackendFramework.Services
             // Get every word with all of its information.
             var allWords = await wordRepo.GetAllWords(projectId);
             var frontier = await wordRepo.GetAllFrontier(projectId);
+            // All words in the frontier with any senses are considered current.
+            // The Combine does not import senseless entries and the interface is supposed to prevent creating them.
             var activeWords = frontier.Where(
                 x => x.Senses.Any(s => s.Accessibility == Status.Active || s.Accessibility == Status.Protected)).ToList();
             var hasFlags = activeWords.Any(w => w.Flag.Active);
@@ -308,11 +310,11 @@ namespace BackendFramework.Services
             // Get all project speakers for exporting audio and consents.
             var projSpeakers = await speakerRepo.GetAllSpeakers(projectId);
 
-            // All words in the frontier with any senses are considered current.
-            // The Combine does not import senseless entries and the interface is supposed to prevent creating them.
-            // So the words found in allWords with no matching guid in activeWords are exported as 'deleted'.
-            var deletedWords = allWords.Where(
-                x => activeWords.All(w => w.Guid != x.Guid)).DistinctBy(w => w.Guid).ToList();
+            // Deleted words found in allWords with no matching guid in activeWords are exported as 'deleted'.
+            var activeWordGuids = activeWords.Select(w => w.Guid).ToHashSet();
+            var deletedWords = allWords
+                .Where(w => w.Accessibility == Status.Deleted && !activeWordGuids.Contains(w.Guid))
+                .DistinctBy(w => w.Guid).ToList();
             var englishSemDoms = await semDomRepo.GetAllSemanticDomainTreeNodes("en") ?? [];
             var semDomNames = englishSemDoms.ToDictionary(x => x.Id, x => x.Name);
 


imnasnainaec:
Now that #4179 is merged (GUIDs are subtype 4), the database update can be done with a `mongosh` one-liner.

**Target set:** every `{projectId, guid}` that `LiftExport` would previously have exported as deleted — i.e. it has a word in `WordsCollection` with at least one sense (matching `GetAllProjectWordsFilter`), and no `FrontierCollection` word with an `Active`/`Protected` sense. For each such guid with no existing `Deleted` word, insert a copy of its most recently modified word with a new `_id` and `accessibility: "Deleted"`.

Dry run — count the words that would be added:

```js
mongosh CombineDatabase --eval '(() => { const k = w => w.projectId + "/" + w.guid.toString("hex"), seen = new Set(); let n = 0; db.FrontierCollection.find({ senses: { $elemMatch: { accessibility: { $in: ["Active", "Protected"] } } } }, { projectId: 1, guid: 1 }).forEach(w => seen.add(k(w))); db.WordsCollection.find({ accessibility: "Deleted", "senses.0": { $exists: true } }, { projectId: 1, guid: 1 }).forEach(w => seen.add(k(w))); db.WordsCollection.find({ accessibility: { $ne: "Deleted" }, "senses.0": { $exists: true } }, { projectId: 1, guid: 1 }).forEach(w => { if (!seen.has(k(w))) { seen.add(k(w)); n++; } }); return n; })()'
```

(Run on QA on database restored from a prod backup less than a week old: 55223 hits.)

Migration:

```js
mongosh CombineDatabase --eval '(() => { const k = w => w.projectId + "/" + w.guid.toString("hex"), seen = new Set(), ops = []; let n = 0; const flush = () => { if (ops.length) { n += db.WordsCollection.bulkWrite(ops.splice(0), { ordered: false }).insertedCount; } }; db.FrontierCollection.find({ senses: { $elemMatch: { accessibility: { $in: ["Active", "Protected"] } } } }, { projectId: 1, guid: 1 }).forEach(w => seen.add(k(w))); db.WordsCollection.find({ accessibility: "Deleted", "senses.0": { $exists: true } }, { projectId: 1, guid: 1 }).forEach(w => seen.add(k(w))); db.WordsCollection.find({ accessibility: { $ne: "Deleted" }, "senses.0": { $exists: true } }).sort({ modified: -1 }).allowDiskUse().forEach(w => { if (seen.has(k(w))) return; seen.add(k(w)); ops.push({ insertOne: { document: { ...w, _id: new ObjectId(), accessibility: "Deleted" } } }); if (ops.length >= 500) flush(); }); flush(); return `inserted ${n} deleted-word copies`; })()'
```

Notes:

- Idempotent: a second run finds a `Deleted` word for every affected guid and inserts nothing.
- Keys on `projectId + guid`, so it is safe across all projects in one pass and does not need per-project iteration.
- `"senses.0": { $exists: true }` mirrors the 0-sense filter in `GetAllProjectWordsFilter`, so the inserted copy is guaranteed to show up in `GetAllWords` (a senseless copy would be invisible to the export and the migration would appear to have done nothing).
- `.sort({ modified: -1 })` picks the newest word per guid, so the exported `<entry id="{vernacular}_{guid}">` uses the last vernacular the word had. `.allowDiskUse()` keeps the blocking sort from hitting the 32 MB limit. If that is a problem on a large deployment, dropping both is harmless — only the `id` attribute of the deleted entry changes, and FieldWorks matches on guid.
- Requires the #4179 guid migration to have run first; `toString("hex")` on a mixed subtype 3/4 collection would key the same logical guid two different ways.

jasonleenaylor (approved):
The guard is sound: `Status.Deleted` is set only in `CreateModifyDeletedWordAction` (`WordService.cs:85-91`), and `DeleteFrontierWord`, `MergeReplaceFrontier` and `RevertMergeReplaceFrontier` all route through it, so nothing that legitimately leaves the frontier is missed.

Non-blocking: the deleted-entry path has no test either side of this change — `LiftServiceTests` never calls `LiftExport`, and Codecov's "modified lines covered" only means they run. A case asserting that an inactive word without `Status.Deleted` produces no deleted entry would pin the behavior this PR exists for.

imnasnainaec (commented):
Test added.

> @imnasnainaec made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/thecombine/4178#-:-P2A4FG35CkGzpkz64_4:b-65kv61)* status: 0 of 2 files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #4278, opened 2026-05-08
Fix bugs caught by DeepWiki

### LINKED ISSUE #4277: Evaluate DeepWiki bugs
https://deepwiki.com/search/look-for-bugs-in-this-repo_ea86c71c-bc7f-4eab-9551-28c8f22d85c7?mode=deep

# DeepWiki Q&A with Code Context for Repository: sillsdev/TheCombine
## Q1
Look for bugs in this repo.
### Answer
The user wants me to look for bugs in the TheCombine repository. Let me start by exploring the codebase to understand its structure and then look for potential bugs. I'll search for common bug patterns like error handling issues, null/undefined checks, race conditions, etc.

Let me do a more targeted search for actual bugs in the codebase. I'll look at several areas simultaneously:
1. Error handling issues
2. Off-by-one errors
3. Race conditions
4. Missing await/async issues
5. Logic errors in reducers
6. Unhandled promise rejections

Let me dig deeper into specific areas that look suspicious. I noticed a few potential bugs:

1. In `MergeDupsReducer.ts` line 311: `state = defaultState;` - In Redux Toolkit's Immer-based reducers, reassigning `state` directly doesn't work as expected. You need to `return defaultState` instead.

2. In `GoalActions.ts` line 104: `initialStep != getState().goalsState.currentGoal.currentStep` - uses `!=` (loose inequality) instead of `!==` (strict inequality).

3. In `asyncSignUp` in `LoginActions.ts` line 100-102: `setTimeout(async () => { dispatch(asyncLogIn(...)) }, 1000)` - the dispatch inside setTimeout is not awaited, and the setTimeout itself is not awaited. This could cause issues.

4. In `CreateProjectActions.ts` lines 27-30 and 49-52: `setTimeout(() => { dispatch(asyncCreateUserEdits()); router.navigate(Path.ProjSettings); }, 1000)` - the dispatch inside setTimeout is not awaited.

Let me look more carefully at these potential bugs and also search for more.

Let me now look at a few more specific potential bugs I've identified:

1. **`setDataAction` in `MergeDupsReducer.ts` line 311**: `state = defaultState;` - In Redux Toolkit's Immer-based reducers, reassigning the `state` parameter directly does NOT work. You need to `return defaultState` instead. This is a real bug - when `action.payload.length === 0`, the state won't actually be reset.

2. **`asyncAdvanceStep` in `GoalActions.ts` line 104**: Uses `!=` (loose inequality) instead of `!==` (strict inequality). This is a style issue but could be a bug if the types differ.

3. **`asyncSignUp` in `LoginActions.ts` line 101**: `dispatch(asyncLogIn(...))` inside `setTimeout` - the dispatch result is not awaited, but since it's inside a setTimeout, this is intentional (fire and forget after 1 second delay). However, the `async` keyword on the setTimeout callback is unnecessary and misleading.

4. **`SaveDeferButtons.tsx` line 26**: Uses `==` (loose equality) instead of `===` (strict equality) to compare `protectedDataOverrideEnabled == OffOnSetting.On`.

Let me look more carefully at these bugs and also check for more.

Let me look at the `setDataAction` bug more carefully. In Redux Toolkit with Immer, if you reassign `state = defaultState`, it doesn't work because Immer tracks mutations to the proxy object. To reset state, you need to `return defaultState`. This is a real bug.

Let me also look at the `==` vs `===` issue in SaveDeferButtons.tsx.

Let me also look at the `asyncSignUp` issue more carefully - the `dispatch(asyncLogIn(...))` call inside setTimeout is not awaited, but since it's inside a setTimeout, the async function wrapping it is unnecessary.

Let me search for more bugs.

Let me now synthesize the bugs I've found from my investigation.

**Bug 1: `setDataAction` in `MergeDupsReducer.ts` - Immer state reassignment doesn't work**

In `src/goals/MergeDuplicates/Redux/MergeDupsReducer.ts` line 311:
```ts
setDataAction: (state, action) => {
  if (action.payload.length === 0) {
    state = defaultState;  // BUG: This does NOT work in Immer!
  } else {
    // ...
  }
},
```

In Redux Toolkit's Immer-based reducers, reassigning the `state` parameter directly has no effect. The Immer proxy tracks mutations to the existing state object. To replace the entire state, you must `return defaultState` instead. The `else` branch correctly mutates `state` directly, but the `if` branch is broken.

Compare with `clearTreeAction` which correctly does `return defaultState`.

**Bug 2: `asyncAdvanceStep` in `GoalActions.ts` - loose inequality `!=` instead of `!==`**

In `src/goals/Redux/GoalActions.ts` line 104:
```ts
if (initialStep != getState().goalsState.currentGoal.currentStep) {
```

This uses `!=` (loose inequality) instead of `!==` (strict inequality). Since `currentStep` is a number, this is unlikely to cause a real bug in practice (numbers compare the same with `==` and `===` unless one is `null`/`undefined`), but it violates the project's own style guide which uses strict equality everywhere else. The grep search confirmed this is the only `!=` in the entire `src/**/*.ts` codebase.

**Bug 3: `asyncSignUp` in `LoginActions.ts` - unawaited dispatch inside setTimeout**

In `src/components/Login/Redux/LoginActions.ts` lines 100-102:
```ts
setTimeout(async () => {
  dispatch(asyncLogIn(username, password, onLoginSuccess));
}, 1000);
```

The `dispatch(asyncLogIn(...))` call returns a Promise (since `asyncLogIn` is a thunk), but it's not awaited. The `async` keyword on the setTimeout callback is also unnecessary and misleading. If the login fails, the error won't be caught. This is a minor issue since the intent is fire-and-forget after a delay, but the `async` keyword is misleading.

**Bug 4: `SaveDeferButtons.tsx` - loose equality `==` instead of `===`**

In `src/goals/MergeDuplicates/MergeDupsStep/SaveDeferButtons.tsx` line 26:
```ts
state.currentProjectState.project.protectedDataOverrideEnabled ==
  OffOnSetting.On
```

Uses `==` (loose equality) instead of `===` (strict equality). Since `OffOnSetting.On` is likely a string enum value, this could work in practice but violates the style guide.

**Bug 5: `LiftHelper.cs` - `IsProtected` trait condition logic**

In `Backend/Helper/LiftHelper.cs` lines 85-86:
```csharp
entry.Traits.Any(t => !t.Value.Equals("stem", StringComparison.OrdinalIgnoreCase) ||
    !t.Name.Replace("-", "").Equals(TraitNames.MorphType, StringComparison.OrdinalIgnoreCase))
```

This condition says: a trait is "protected" if its value is NOT "stem" OR its name is NOT "morphtype". By De Morgan's law, this is equivalent to: NOT (value == "stem" AND name == "morphtype"). So a trait is unprotected only if BOTH its value is "stem" AND its name is "morphtype". This seems correct - the only unprotected trait is morph-type "stem".

Actually wait, let me re-read this. The `Any` returns true if ANY trait satisfies the condition. The condition for a single trait `t` is:
- `!t.Value.Equals("stem")` OR `!t.Name.Replace("-", "").Equals(TraitNames.MorphType)`

This means: a trait triggers protection if it's NOT (value="stem" AND name="morphtype"). So any trait that isn't morph-type "stem" will trigger protection. This seems correct.

But wait - what about a trait with name="morphtype" and value="stem"? The condition would be:
- `!true || !true` = `false || false` = `false`

So this trait does NOT trigger protection. That's correct.

What about a trait with name="morphtype" and value="root"?
- `!false || !true` = `true || false` = `true`

So this trait DOES trigger protection. That's correct.

What about a trait with name="dialect-labels" and value="stem"?
- `!true || !false` = `false || true` = `true`

So this trait DOES trigger protection. That's correct.

OK so the logic seems correct. Let me not flag this as a bug.

**Bug 6: `LiftService.cs` - potential `IndexOutOfRangeException` when splitting semantic domain string**

In `Backend/Services/LiftService.cs` line 891-898:
```csharp
var splitSemDom = semanticDomainString.Split(" ", 2);
newSense.SemanticDomains.Add(
    new SemanticDomain
    {
        Id = splitSemDom[0],
        MongoId = ObjectId.GenerateNewId().ToString(),
        Name = splitSemDom[1].Trim().Normalize(FormC)  // Could throw if no space in string
    });
```

If `semanticDomainString` doesn't contain a space (e.g., it's just a number like "1.1" with no name), then `splitSemDom[1]` would throw an `IndexOutOfRangeException`. This is a real potential bug.

Let me now summarize the bugs I've found clearly.

Actually, let me reconsider the `setDataAction` bug more carefully. In Redux Toolkit with Immer:
- If you mutate the `state` proxy, Immer applies those mutations.
- If you `return` a new value, Immer uses that as the new state.
- If you reassign the `state` parameter (like `state = defaultState`), this does NOT work because you're just reassigning the local variable, not the actual state.

So `state = defaultState` on line 311 is indeed a bug - when `action.payload.length === 0`, the state will NOT be reset to `defaultState`. The correct fix is `return defaultState`.

This is confirmed by comparing with `clearTreeAction` at line 40-42 which correctly uses `return defaultState`.

Let me now present these bugs clearly.

Here are the bugs I found across the codebase:

---

## Bug 1: Immer state reassignment has no effect (Redux Toolkit)

**File:** `src/goals/MergeDuplicates/Redux/MergeDupsReducer.ts`, line 311

```ts
setDataAction: (state, action) => {
  if (action.payload.length === 0) {
    state = defaultState;  // BUG: no-op in Immer
  } else { ... }
},
```

In Redux Toolkit's Immer-based reducers, reassigning the `state` parameter is a local variable reassignment — it does **not** update the store. The correct pattern is `return defaultState`. When an empty word list is dispatched, the state silently stays unchanged instead of resetting.

Compare with `clearTreeAction` in the same file, which correctly uses `return defaultState`: TheCombine:40-42 

The broken branch: TheCombine:309-312 

---

## Bug 2: Loose inequality `!=` instead of `!==`

**File:** `src/goals/Redux/GoalActions.ts`, line 104

```ts
if (initialStep != getState().goalsState.currentGoal.currentStep) {
```

This is the **only** use of `!=` in the entire `src/**/*.ts` codebase (confirmed by grep). While it works for numbers in practice, it violates the project's strict equality convention and could silently misbehave if `currentStep` is ever `null` or `undefined` (e.g., `0 != null` is `true` with `!=` but `0 !== null` is also `true` — however `null != undefined` is `false` while `null !== undefined` is `true`). TheCombine:100-115 

---

## Bug 3: Potential `IndexOutOfRangeException` in LIFT import

**File:** `Backend/Services/LiftService.cs`, lines 891–898

```csharp
var splitSemDom = semanticDomainString.Split(" ", 2);
newSense.SemanticDomains.Add(new SemanticDomain {
    Id = splitSemDom[0],
    Name = splitSemDom[1].Trim()...  // throws if no space in string
});
```

If a semantic domain string contains no space (e.g., just `"1.1"` with no name), `splitSemDom[1]` throws an `IndexOutOfRangeException`. There is no guard checking `splitSemDom.Length > 1` before accessing index 1. TheCombine:891-898 

---

## Bug 4: Loose equality `==` in component selector

**File:** `src/goals/MergeDuplicates/MergeDupsStep/SaveDeferButtons.tsx`, line 26

```ts
state.currentProjectState.project.protectedDataOverrideEnabled ==
  OffOnSetting.On
```

Uses `==` (loose equality) instead of `===`. Since `OffOnSetting` is a string enum, this is unlikely to cause a runtime bug today, but it is inconsistent with the rest of the codebase and could mask type mismatches. TheCombine:23-28 

---

## Bug 5: Unawaited async thunk dispatch inside `setTimeout`

**File:** `src/components/Login/Redux/LoginActions.ts`, lines 100–102

```ts
setTimeout(async () => {
  dispatch(asyncLogIn(username, password, onLoginSuccess));
}, 1000);
```

`dispatch(asyncLogIn(...))` returns a `Promise` that is not awaited. The `async` keyword on the callback is misleading and serves no purpose. If the login thunk throws, the error is silently swallowed. The same pattern appears in `src/components/ProjectScreen/CreateProjectActions.ts` lines 27–30 and 49–52. TheCombine:100-102 TheCombine:27-30 

---

## Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **High** | `MergeDupsReducer.ts:311` | `state = defaultState` is a no-op in Immer; should be `return defaultState` |
| 2 | Medium | `GoalActions.ts:104` | `!=` loose inequality; should be `!==` |
| 3 | Medium | `LiftService.cs:897` | `splitSemDom[1]` throws if semantic domain string has no space |
| 4 | Low | `SaveDeferButtons.tsx:26` | `==` loose equality; should be `===` |
| 5 | Low | `LoginActions.ts:101` | Unawaited promise from async thunk in `setTimeout` |
### Citations
**File:** src/goals/MergeDuplicates/Redux/MergeDupsReducer.ts (L40-42)
```typescript
    clearTreeAction: () => {
      return defaultState;
    },
```
**File:** src/goals/MergeDuplicates/Redux/MergeDupsReducer.ts (L309-312)
```typescript
    setDataAction: (state, action) => {
      if (action.payload.length === 0) {
        state = defaultState;
      } else {
```
**File:** src/goals/Redux/GoalActions.ts (L100-115)
```typescript
export function asyncAdvanceStep() {
  return async (dispatch: StoreStateDispatch, getState: () => StoreState) => {
    const initialStep = getState().goalsState.currentGoal.currentStep;
    dispatch(incrementGoalStep());
    if (initialStep != getState().goalsState.currentGoal.currentStep) {
      // Update data.
      dispatch(updateStepFromData());
      // Dispatch to state.
      const updatedGoal = getState().goalsState.currentGoal;
      dispatch(dispatchStepData(updatedGoal));
      // Save to database.
      await saveCurrentStep(updatedGoal);
    } else {
      goalCleanup(getState().goalsState.currentGoal);
    }
  };
```
**File:** Backend/Services/LiftService.cs (L891-898)
```csharp
                        var splitSemDom = semanticDomainString.Split(" ", 2);
                        newSense.SemanticDomains.Add(
                            new SemanticDomain
                            {
                                Id = splitSemDom[0],
                                MongoId = ObjectId.GenerateNewId().ToString(),
                                Name = splitSemDom[1].Trim().Normalize(FormC)
                            });
```
**File:** src/goals/MergeDuplicates/MergeDupsStep/SaveDeferButtons.tsx (L23-28)
```typescript
  const hasProtected = useAppSelector(
    (state: StoreState) =>
      state.mergeDuplicateGoal.hasProtected &&
      state.currentProjectState.project.protectedDataOverrideEnabled ==
        OffOnSetting.On
  );
```
**File:** src/components/Login/Redux/LoginActions.ts (L100-102)
```typescript
        setTimeout(async () => {
          dispatch(asyncLogIn(username, password, onLoginSuccess));
        }, 1000);
```
**File:** src/components/ProjectScreen/CreateProjectActions.ts (L27-30)
```typescript
    setTimeout(() => {
      dispatch(asyncCreateUserEdits());
      router.navigate(Path.ProjSettings);
    }, 1000);
```

imnasnainaec:
Fixes #4277

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4278)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential crash when processing semantic-domain data with incomplete formatting.
  * Improved state reset handling during data merging operations.
  * Corrected step-change detection logic to prevent unintended state updates.

* **Code Quality**
  * Enhanced comparison logic consistency across multiple components for better reliability.

### DIFF
--- a/Backend/Services/LiftService.cs
+++ b/Backend/Services/LiftService.cs
@@ -894,7 +894,7 @@ namespace BackendFramework.Services
                             {
                                 Id = splitSemDom[0],
                                 MongoId = ObjectId.GenerateNewId().ToString(),
-                                Name = splitSemDom[1].Trim().Normalize(FormC)
+                                Name = splitSemDom.Length > 1 ? splitSemDom[1].Trim().Normalize(FormC) : ""
                             });
                     }
 
--- a/src/components/Login/Redux/LoginActions.ts
+++ b/src/components/Login/Redux/LoginActions.ts
@@ -97,7 +97,7 @@ export function asyncSignUp(
       .then(() => {
         dispatch(signupSuccess());
         onSignupSuccess?.();
-        setTimeout(async () => {
+        setTimeout(() => {
           dispatch(asyncLogIn(username, password, onLoginSuccess));
         }, 1000);
       })
--- a/src/goals/MergeDuplicates/MergeDupsStep/SaveDeferButtons.tsx
+++ b/src/goals/MergeDuplicates/MergeDupsStep/SaveDeferButtons.tsx
@@ -23,7 +23,7 @@ export default function SaveDeferButtons(): ReactElement {
   const hasProtected = useAppSelector(
     (state: StoreState) =>
       state.mergeDuplicateGoal.hasProtected &&
-      state.currentProjectState.project.protectedDataOverrideEnabled ==
+      state.currentProjectState.project.protectedDataOverrideEnabled ===
         OffOnSetting.On
   );
   const overrideProtection = useAppSelector(
--- a/src/goals/MergeDuplicates/Redux/MergeDupsReducer.ts
+++ b/src/goals/MergeDuplicates/Redux/MergeDupsReducer.ts
@@ -308,7 +308,7 @@ const mergeDuplicatesSlice = createSlice({
 
     setDataAction: (state, action) => {
       if (action.payload.length === 0) {
-        state = defaultState;
+        return defaultState;
       } else {
         const words: Hash<Word> = {};
         const senses: Hash<MergeTreeSense> = {};
--- a/src/goals/Redux/GoalActions.ts
+++ b/src/goals/Redux/GoalActions.ts
@@ -101,7 +101,7 @@ export function asyncAdvanceStep() {
   return async (dispatch: StoreStateDispatch, getState: () => StoreState) => {
     const initialStep = getState().goalsState.currentGoal.currentStep;
     dispatch(incrementGoalStep());
-    if (initialStep != getState().goalsState.currentGoal.currentStep) {
+    if (initialStep !== getState().goalsState.currentGoal.currentStep) {
       // Update data.
       dispatch(updateStepFromData());
       // Dispatch to state.


jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4278#-:-Os7lXypACapnPIrgHTQ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #4323, opened 2026-07-06
Add timeout and retry to combine_restore kubectl cp

### LINKED ISSUE #4322: combine_restore.py hangs forever when kubectl cp stalls (no timeout/retry)
## Summary

`maintenance/scripts/combine_restore.py` copies backend files into the backend pod with a per-project `kubectl cp` loop. `kubectl cp` streams a tar over the exec channel and can stall intermittently. Because `combine_app.kubectl()` → `maint_utils.run_cmd()` calls `subprocess.run(...)` with **no `timeout`**, a stalled copy hangs **forever** with no error and no output. This path is The Combine's disaster-recovery restore, so a silent infinite hang is a serious reliability problem.

## Observed

During a prod→qa restore, the backend-file copy printed `DEBUG:Copying <projectId> ...` and then produced no further output for 30+ minutes. The target PVC had plenty of headroom (26% space used, 1% inodes), so it was **not** `ENOSPC`. Aborting and re-running stalled again — on a **different** project — indicating an intermittent `kubectl cp` stream stall rather than a bad/oversized project.

## Root cause

- `maintenance/scripts/maint_utils.py` `run_cmd()` runs `subprocess.run(...)` with no `timeout`.
- `maintenance/scripts/combine_app.py` `kubectl()` / `exec()` inherit this — no timeout.
- The restore's backend-file loop (and the DB `kubectl cp`, ~L163) can therefore hang indefinitely on any stalled stream:

```python
# combine_restore.py — backend-file loop
for item in os.listdir(backend_files_subdir):
    if item.startswith(".") or item == "lost+found":
        continue
    local_item = os.path.join(backend_files_subdir, item)
    combine.kubectl(["cp", local_item, remote_subdir, "--no-preserve"])
```

## Proposed fix

- Add a `timeout` parameter to `run_cmd()` (and thread it through `kubectl()`/`exec()`); apply a sensible timeout to the copy calls.
- Wrap each copy in a bounded retry so a transient stall is killed and retried instead of hanging.
- On repeated failure, log the failing project and exit non-zero so it surfaces as an error rather than a silent hang.
- Optionally replace the per-project `kubectl cp` loop with a single tar-pipe (`tar cf - … | kubectl exec -i … -- tar xf - …`), which is more robust for large transfers — still guarded by a timeout/retry.

## Impact

`combine_backup.py` / `combine_restore.py` are the prod backup/restore (disaster recovery). A restore that can silently hang undermines that guarantee.

imnasnainaec:
## Fixes

Fixes #4322

## Problem

`maintenance/scripts/combine_restore.py` copies backend files into the backend pod with a per-project `kubectl cp` loop (and copies the DB dump the same way). `kubectl cp` streams a tar over the exec channel and can stall intermittently. Because `maint_utils.run_cmd()` called `subprocess.run(...)` with **no `timeout`**, a stalled copy hung **forever** with no error and no output. Since this is The Combine's disaster-recovery restore, a silent infinite hang is a serious reliability problem.

## Changes

- **`maint_utils.py`**: add an optional `timeout` parameter to `run_cmd()`, passed through to `subprocess.run(...)`. When set, a stalled command is killed and `subprocess.TimeoutExpired` is raised for the caller to handle. Default `None` preserves existing behavior.
- **`combine_app.py`**: thread `timeout` through `CombineApp.exec()` and `CombineApp.kubectl()` (also adding `check_results` to `kubectl()`).
- **`combine_restore.py`**: wrap each `kubectl cp` (the database dump and every backend-file item) in a new `kubectl_cp()` helper that:
  - bounds each attempt with a timeout so a stalled stream is killed;
  - retries a few times so a transient stall is recovered instead of aborting the whole restore;
  - on repeated failure, logs the failing item (project) and exits non-zero so it surfaces as an error rather than a silent hang.
- The timeout and retry count are configurable via the `kubectl_cp_timeout` (default 300s) and `kubectl_cp_retries` (default 3) environment variables.

I kept the existing per-item `kubectl cp` loop rather than switching to a single `tar`-pipe (listed as optional in the issue) to keep the change focused; the timeout/retry guard applies either way.

## Testing

- `tox` equivalents pass locally on the changed files: `isort --check`, `black --check`, `flake8`, and `mypy --platform linux maintenance/scripts` (strict) — all clean.
- Exercised the new logic directly:
  - `run_cmd(..., timeout=1)` on a 30s sleep raises `TimeoutExpired` (no longer hangs) and still returns output for fast commands.
  - `kubectl_cp()` succeeds without retry on a clean copy (and passes `check_results=False` + `timeout`), recovers from transient stalls (timeout → timeout → success), and exits non-zero after exhausting retries on repeated failure.

Devin review: https://app.devin.ai/review/sillsdev/TheCombine/pull/4323

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4323)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backup and restore reliability by retrying file copy operations that can stall or fail temporarily.
  * Added timeouts to long-running copy and command operations so they fail more predictably instead of hanging indefinitely.
  * Backup and restore flows now use clearer copy-step labeling for better troubleshooting.

* **Chores**
  * Added configurable environment-based settings for copy timeouts and retry attempts.

### DIFF
--- a/maintenance/scripts/combine_app.py
+++ b/maintenance/scripts/combine_app.py
@@ -17,8 +17,9 @@ from maint_utils import run_cmd
 # A `kubectl cp` streams a tar over the exec channel and can stall intermittently,
 # which would otherwise hang forever.  Bound each copy with a timeout so a stalled
 # stream gets killed, and retry a few times so a transient stall is recovered.
-CP_ATTEMPTS = int(os.getenv("kubectl_cp_attempts", "3"))
-CP_TIMEOUT = float(os.getenv("kubectl_cp_timeout", "300"))
+# Clamp to a sane floor so a misconfigured env var can't skip the copy entirely.
+CP_ATTEMPTS = max(1, int(os.getenv("kubectl_cp_attempts", "3")))
+CP_TIMEOUT = max(10.0, float(os.getenv("kubectl_cp_timeout", "300")))
 # A fast-failing copy can exhaust all attempts within the same second, giving a
 # transient failure no time to clear; pause between attempts so the retries span time.
 CP_RETRY_DELAY = 3.0
--- a/maintenance/scripts/combine_app.py
+++ b/maintenance/scripts/combine_app.py
@@ -39,6 +39,7 @@ class CombineApp:
         *,
         exec_opts: Optional[List[str]] = None,
         check_results: bool = True,
+        timeout: Optional[float] = None,
     ) -> subprocess.CompletedProcess[str]:
         """
         Run a kubectl 'exec' command in a Combine Kubernetes cluster.
@@ -52,6 +53,8 @@ class CombineApp:
                      command, for example, to specify a working directory or a
                      specific user to run the command.
             check_results: Indicate if subprocess should not check for failure.
+            timeout: If set, kill the command and raise subprocess.TimeoutExpired
+                     when it runs longer than this many seconds.
         Returns a subprocess.CompletedProcess.
         """
         exec_opts = exec_opts or []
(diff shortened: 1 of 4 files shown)


jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4323#-:-Ox1z1L25e-SphOWilY-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #4209, opened 2026-03-19
Remove one-shot db guid migration script

imnasnainaec:
Partially revert #4179, removing things only needed for generating the script to migrate GUIDs in the database from subtype 3 to 4.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4209)

## Summary by CodeRabbit

* **Chores**
  * Removed database migration script and related TypeScript configuration
  * Removed BSON development dependency
  * Updated `.gitignore` to include JavaScript files in the database directory

PR #4327, opened 2026-07-09
Cap UserEdit edits to fix 16 MB limit (do NOT merge until after v3)

### LINKED ISSUE #4320: UserEdit document can exceed MongoDB's 16 MB limit, permanently blocking goal/step writes
**Describe the bug**

Once a user's `UserEdit` document for a project grows close to MongoDB's hard
16 MB per-document limit, every subsequent goal/step write to it fails with a 500,
and the user is permanently unable to record new work in that project.

Reads still succeed, so the document is not corrupt — the failure is purely on write.

Observed in production (Honeycomb):

- Route: `POST v1/projects/{projectId}/useredits/{userEditId}`
- `http.response.status_code: 500`
- `error.type: MongoDB.Driver.MongoWriteException`
- `exception.message`: `WriteError { Code: 17419, Message: "Resulting document after
  update is larger than 16777216" }`
- Stack: `UserEditRepository.Replace (line 94)` ← `UserEditService.AddGoalToUserEdit
  (line 55)` ← `UserEditController.UpdateUserEditGoal (line 136)`

The affected `UserEdit` had 500+ entries in its `edits` array, sitting just under 16 MB;
the payload to add was only 78 bytes.

**Root cause**

Two design factors combine:

1. **The `edits` array is unbounded** (`UserEdit.Edits`), and each `Edit` carries a
   `changes` JSON blob plus `stepData`. It grows with every goal a user completes and
   is never trimmed.
2. **Writes rewrite the entire document.** `AddGoalToUserEdit`, `AddStepToGoal`, and
   `UpdateStepInGoal` all load the full `UserEdit`, mutate the array in memory, and call
   `UserEditRepository.Replace` → `ReplaceOneAsync` with the whole document. When the
   already-large document plus one more edit crosses 16 MB, Mongo rejects the write.

Because it's a full-document replace, there is no way for the user to make further
progress in that project once the threshold is reached — the array can only grow.

**To Reproduce**

1. As a single user, accumulate a very large number of goal edits in one project
   (or seed a `UserEdit` whose `edits` array approaches 16 MB).
2. Attempt to start/advance another goal (POST to `.../useredits/{userEditId}`) or a
   step (PUT to the same route).
3. Backend returns 500; Honeycomb shows `MongoWriteException` code 17419.

**Expected behavior**

Recording goals/steps should not fail as edit history accumulates. The stored history
should be bounded (or stored such that no single document can hit the 16 MB limit), and
writes should not require rewriting the whole document.

**Proposed fixes** (for discussion)

- Short term: replace the append-then-`Replace` pattern with an atomic `$push` +
  `$slice` that caps the `edits` array length server-side, bounding growth going forward.
- Consider trimming/limiting what is stored in each `Edit.changes`.
- Longer term: move edits to their own collection (one document per edit) so no single
  document is bounded by the 16 MB limit. (Schema migration.)

**Note on already-affected data**

At least one production `UserEdit` is already at the limit and needs its `edits` array
trimmed in place to unblock the user. Trim in place — do not delete the document, since
the user's `WorkedProjects` map references its `_id`.

**Environment**

- Component: Backend (`UserEditController` / `UserEditService` / `UserEditRepository`)
- Server: production (`thecombine.app`)
- Database: MongoDB (16 MB BSON document limit)

imnasnainaec:
> [!CAUTION]
> **DO NOT MERGE until after the v3 release.** This PR changes the `UserEdit` write path and includes a data migration that trims production data. It is intentionally held back so it doesn't land in the v3 release; merge only once v3 has shipped and the migration plan below has been agreed on.

Resolves #4320

## Problem

A `UserEdit` document's `edits` array is unbounded, and every goal/step write loaded the full document and rewrote it with `ReplaceOneAsync`. Once a document neared MongoDB's 16 MB limit, every write failed with `WriteError 17419` and the user was permanently blocked from recording goal/step progress in that project (observed in production; see #4320).

## Fix

- **Targeted atomic updates instead of full-document replaces.** `UserEditRepository.Replace` is removed; goal and step writes now use `$push`/`$set` on just the affected `edits` element.
- **Server-side cap on `edits`.** Appending a new edit uses `$push` + `$slice: -UserEdit.MaxEdits`, so the array is atomically trimmed to the most recent `MaxEdits` entries on every append. Growth is bounded going forward.
- `MaxEdits = 250`: production documents that hit the limit held 387–976 edits at 14–16 MB (up to ~40 kB per edit), so 250 keeps worst-case documents around 10 MB. Value is easy to tune if we want to keep more history.

Because the capped `$push` produces a *smaller* resulting document, deploying this backend also unblocks already-affected users on their next goal write — the migration below is still needed to shrink oversized documents proactively (they slow down reads such as `GetProjectUserEdits`).

Frontend requires no changes: edits are addressed by GUID everywhere; nothing persists an index into the `edits` array.

## Mongo migration instructions

`database/trim-user-edits.js` trims every `UserEdit` document with more than 250 edits down to its most recent 250, in place (documents are never deleted — each user's `workedProjects` references the document `_id`). Run it once against production (and qa) after this PR is deployed:

1. **Back up first** — trimmed entries are unrecoverable without a backup:

   ```bash
   # full backup via the maintenance pod, or at minimum:
   mongodump --db=CombineDatabase --collection=UserEditsCollection
   ```

2. **Run the script** on the database pod:

   ```bash
   kubectl -n thecombine cp database/trim-user-edits.js <database-pod>:/tmp/trim-user-edits.js
   kubectl -n thecombine exec <database-pod> -- mongosh CombineDatabase /tmp/trim-user-edits.js
   ```

   The script prints each affected document (`_id`, project, edit count, size) before trimming and verifies afterward that no document exceeds the cap.

3. Per the precedent of the GUID-subtype migration (#4179 / #4209), the script can be removed from the repo in a follow-up once it has been run everywhere.

## Testing

- `dotnet build BackendFramework.sln`: 0 warnings, 0 errors.
- `dotnet test Backend.Tests`: 1186/1186 passed, including 12 new `UserEditRepositoryTests` integration tests that run the new atomic updates against a real (ephemeral) mongod — append, cap-trim (oldest dropped, newest last), replace-by-guid, step append/overwrite, and false-on-missing cases — plus a controller-level trim test via the mock.
- `npm run fmt-backend`: clean.
- Step writes previously stamped `Edit.Modified`; the new `$push`/`$set` updates preserve that (the goal history UI displays it).

## Notes for reviewers

- `Replace` had no callers outside `UserEditService`, so it was removed from `IUserEditRepository`, the repository, and the mock rather than left as dead code that reintroduces the full-document-rewrite pattern.
- Each `Edit.changes` blob is still unbounded (a single huge edit could in theory exceed limits on its own); trimming what's stored per edit and/or moving edits to their own collection remain the longer-term options discussed in #4320.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4327)

### DIFF
--- a/Backend.Tests/Controllers/UserEditControllerTests.cs
+++ b/Backend.Tests/Controllers/UserEditControllerTests.cs
@@ -166,6 +166,28 @@ namespace Backend.Tests.Controllers
         }
 
         [Test]
+        public async Task TestAddGoalToUserEditTrimsToMaxEdits()
+        {
+            var userEdit = new UserEdit { ProjectId = ProjId };
+            foreach (var _ in Range(0, UserEdit.MaxEdits))
+            {
+                userEdit.Edits.Add(new Edit());
+            }
+            userEdit = await _userEditRepo.Create(userEdit);
+            var oldestEditGuid = userEdit.Edits.First().Guid;
+
+            var newEdit = new Edit();
+            var result = await _userEditController.UpdateUserEditGoal(ProjId, userEdit.Id, newEdit);
+            Assert.That(result, Is.InstanceOf<OkResult>());
+
+            var repoUserEdit = await _userEditRepo.GetUserEdit(ProjId, userEdit.Id);
+            Assert.That(repoUserEdit, Is.Not.Null);
+            Assert.That(repoUserEdit.Edits, Has.Count.EqualTo(UserEdit.MaxEdits));
+            Assert.That(repoUserEdit.Edits.Last().Guid, Is.EqualTo(newEdit.Guid));
+            Assert.That(repoUserEdit.Edits.Select(e => e.Guid), Does.Not.Contain(oldestEditGuid));
+        }
+
+        [Test]
         public async Task TestAddStepToGoal()
         {
             // Generate db entry to test.
--- a/Backend.Tests/Mocks/UserEditRepositoryMock.cs
+++ b/Backend.Tests/Mocks/UserEditRepositoryMock.cs
@@ -54,11 +54,59 @@ namespace Backend.Tests.Mocks
             return Task.FromResult(rmCount > 0);
         }
 
-        public Task<bool> Replace(string projectId, string userEditId, UserEdit userEdit)
+        public Task<bool> AddEdit(string projectId, string userEditId, Edit edit)
         {
-            var rmCount = _userEdits.RemoveAll(ue => ue.Id == userEditId);
-            _userEdits.Add(userEdit);
-            return Task.FromResult(rmCount > 0);
+            var userEdit = _userEdits.Find(ue => ue.ProjectId == projectId && ue.Id == userEditId);
+            if (userEdit is null)
+            {
+                return Task.FromResult(false);
+            }
+            userEdit.Edits.Add(edit.Clone());
+            // Mirror the real repository's negative $slice: keep only the most recent MaxEdits entries.
+            if (userEdit.Edits.Count > UserEdit.MaxEdits)
+            {
+                userEdit.Edits.RemoveRange(0, userEdit.Edits.Count - UserEdit.MaxEdits);
+            }
+            return Task.FromResult(true);
+        }
+
+        public Task<bool> ReplaceEdit(string projectId, string userEditId, Edit edit)
+        {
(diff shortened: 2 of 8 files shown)


imnasnainaec:
Superseded by a rework that will move edits into their own collection (one document per edit) instead of capping the array.

PR #4347, opened 2026-07-27
Prep backend for .NET 10: replace obsolete hosting/crypto APIs

imnasnainaec:
## Summary

Prep work toward #4254 (Update .NET from 8 to 10). Replaces two APIs that are already obsolete-marked as of
.NET 9/10 and would fail the build outright once the TargetFramework is bumped, since the project builds with
`TreatWarningsAsErrors=true`. Neither replacement requires .NET 10, so landing them separately shrinks the scope of
the eventual TFM-bump PR.

No behavior changes: same password hashes, same service registrations, same middleware pipeline, same log messages.

## `Backend/Helper/PasswordHash.cs` — `SYSLIB0060`

The `Rfc2898DeriveBytes` constructor is obsolete. Replaced with the static `Rfc2898DeriveBytes.Pbkdf2(...)` method,
passing the password as UTF-8 bytes. The old constructor also used UTF-8 internally, so existing stored hashes
continue to validate.

## `Backend/Program.cs` — `ASPDEPR008`

`WebHost.CreateDefaultBuilder(...).UseStartup<Startup>()` is obsolete. `Main` now uses `WebApplicationBuilder`
minimal hosting and invokes `Startup`'s two methods directly:

```csharp
var builder = WebApplication.CreateBuilder(args);
var startup = new Startup(builder.Configuration);
startup.ConfigureServices(builder.Services);
var app = builder.Build();
startup.Configure(app, app.Environment, app.Lifetime);
app.Run();
```

`CreateWebHostBuilder` is removed; it had no other callers.

## `Backend/Startup.cs`

The old hosting model injected `ILogger<Startup>` into `Startup`'s constructor, resolving it from the host's
internal service provider. Minimal hosting has no equivalent provider, and one can't be conjured before
`builder.Build()` without building a throwaway container. Rather than do that, each logger use now gets the real
application logger at the point where one is actually available:

- **Constructor** takes only `IConfiguration`; the `_logger` field is gone.
- **`Settings` configuration** (the only pre-`Build()` code path that logs) moves from
  `services.Configure<Settings>(options => ...)` to
  `services.AddOptions<Settings>().Configure<ILogger<Startup>>((options, logger) => ...)`. The callback already ran
  lazily — on first `IOptions<Startup.Settings>` resolution, after the app is built — so this only changes where its
  logger comes from. `CheckedEnvironmentVariable` and `CheckedEnvironmentVariablePositiveInt` are now `static` and
  take the logger as a parameter.
- **`Configure(...)`** resolves `ILogger<Startup>` from `app.ApplicationServices` and passes it to
  `CreateAdminUser`.
- **The JWT secret key check** is the one message emitted during `ConfigureServices`, where no logging pipeline
  exists yet. It writes to stderr and then throws, as before.

Every consumer of `Startup.Settings` injects `IOptions<Startup.Settings>` (singleton), not `IOptionsSnapshot` or
`IOptionsMonitor`, so the configure callback still runs exactly once and the environment-variable warnings are not
repeated per scope.

## Verification

- `dotnet build` — clean, 0 warnings.
- `dotnet test` — 1175 passed.
- `npm run fmt-backend-check` — clean.
- Manual run: the environment-variable messages and `No admin user name provided, skipped admin creation` appear as
  `BackendFramework.Startup` records through the application's logging pipeline; unsetting `COMBINE_JWT_SECRET_KEY`
  prints the stderr message before the exception. Neither path is covered by unit tests.

Created by Claude Sonnet 5 and Claude Opus 5

---

Devin: https://app.devin.ai/review/sillsdev/TheCombine/pull/4347

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4347)

### DIFF
--- a/Backend/Program.cs
+++ b/Backend/Program.cs
@@ -1,7 +1,5 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 
 namespace BackendFramework
 {
@@ -11,34 +9,11 @@ namespace BackendFramework
         public static void Main(string[] args)
         {
             var builder = WebApplication.CreateBuilder(args);
-
-            WebApplication app;
-            using (var startupServices = CreateStartupServiceProvider(builder.Services))
-            {
-                var startup = new Startup(
-                    startupServices.GetRequiredService<ILoggerFactory>().CreateLogger<Startup>(),
-                    builder.Configuration);
-                startup.ConfigureServices(builder.Services);
-
-                app = builder.Build();
-                startup.Configure(app, app.Environment, app.Lifetime);
-            }
-
+            var startup = new Startup(builder.Configuration);
+            startup.ConfigureServices(builder.Services);
+            var app = builder.Build();
+            startup.Configure(app, app.Environment, app.Lifetime);
             app.Run();
-        }
-
-        /// <summary>
-        /// Build a service provider from a copy of <paramref name="services"/>, so that <see cref="Startup"/>'s
-        /// bootstrap logger shares the same providers as the app. The caller must dispose the returned provider.
-        /// </summary>
-        private static ServiceProvider CreateStartupServiceProvider(IServiceCollection services)
-        {
-            IServiceCollection startupServices = new ServiceCollection();
-            foreach (ServiceDescriptor service in services)
-            {
-                startupServices.Add(service);
-            }
-            return startupServices.BuildServiceProvider();
         }
     }
 }
--- a/Backend/Program.cs
+++ b/Backend/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace BackendFramework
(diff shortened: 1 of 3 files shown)


Backend/Program.cs
```diff
@@ -9,11 +9,16 @@ public class Program
     {
         public static void Main(string[] args)
         {
-            CreateWebHostBuilder(args).Build().Run();
-        }
+            var builder = WebApplication.CreateBuilder(args);
+
+            using var startupLoggerFactory = LoggerFactory.Create(logging =>
+                logging.AddConfiguration(builder.Configuration.GetSection("Logging")).AddConsole());
+            var startup = new Startup(startupLoggerFactory.CreateLogger<Startup>(), builder.Configuration);
```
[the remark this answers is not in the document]
imnasnainaec: Done.

Backend/Program.cs
```diff
@@ -9,11 +10,35 @@ public class Program
...
         public static void Main(string[] args)
         {
-            CreateWebHostBuilder(args).Build().Run();
+            var builder = WebApplication.CreateBuilder(args);
+
+            WebApplication app;
+            using (var startupServices = CreateStartupServiceProvider(builder.Services))
+            {
+                var startup = new Startup(
+                    startupServices.GetRequiredService<ILoggerFactory>().CreateLogger<Startup>(),
```
hahn-kev: taking a look at where this is used you could eliminate it with just a slight tweak. In Startup there's really only 2 places this is used before the Build is called today. One is logging an error right before throwing an exception when there's no jwt secret, I'd just change that to a console log and then throw. The other one is in `CheckedEnvironmentVariable` which I would modify to take the logger as a parameter, and then change 
```
services.Configure<Settings>(options => ...
```
 to 
```
services.AddOptions<Settings>().Configure<ILogger<Startup>>((options, logger) => ...
```
then pass the logger in to each method (there's also better ways to handle both the parsing and throwing of invalid config but I wouldn't touch that here).

Then all you need to do is tweak `Startup.Configure` to get it's logger from `app.ApplicationServices`.

It's a little bigger of a change to Startup, but it removes the hack of copying the service provider in order to get a logger at startup

hahn-kev (approved):
It looks good, though I do have a suggestion.

Backend/Program.cs
imnasnainaec: Wonderful suggestion, thank you. Done.

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages, resolved 1 discussion, and dismissed @hahn-kev from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4347)* status: 0 of 3 files reviewed, all discussions resolved (waiting on hahn-kev).

jasonleenaylor (approved):
On the strength of Kevin's approval :lgtm:

> @jasonleenaylor reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4347#-:-OydwjTVDJjO5zyqtOdO:bauxz8h)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on hahn-kev).

PR #4351, opened 2026-08-17
Make the installer timeout option work as documented

imnasnainaec:
The documented `timeout` option did not work: `SETUP_OPTS` was set inside `install-base-charts` but used by `install-the-combine`, so runs beginning at `Install-combine` — `update`, `start-at` — dropped it and Helm fell back to its 5 minute default. Fixing that exposed several more ways the option failed.

## Changes

- **Apply it on resumed installs** — compute `SETUP_OPTS` once, after parsing.
- **Remember it for the whole installation**, in `${CONFIG_DIR}/install-timeout`
  - every Helm command runs *after* the restart that `Pre-reqs` usually requires, so a timeout given on the first invocation was discarded before anything could use it
  - the command line overrides and updates the record; `clean` and reaching `Done` discard it; `restart` keeps it
- **Check option values before applying any option**
  - `update timeout bogus` recorded the `Install-combine` state and *then* rejected the value, so the next run skipped `Base-charts`
  - `timeout` or `start-at` with no value exited silently; a non-duration value swallowed the next option
  - both checks moved into `check-opt-value`, not copied
- **Reject unusable timeouts** — non-durations, and anything amounting to zero (`0`, `0s`, `0h0m0s`), which Helm treats as an already expired deadline.
- **Ignore `HELM_TIMEOUT` from the environment** — it was never initialized, so it skipped the format check and would have been recorded.
- **Document it** — the `timeout` and `clean` rows of the installer README.

## Behavior

| Input | Before | After |
| --- | --- | --- |
| `timeout 30m0s v3.0.0`, reboot, rerun | dropped at the reboot | reused until the install finishes |
| `update timeout 30m0s` | silently ignored | applied |
| `update timeout bogus` | state advanced, `Base-charts` skipped | error, saved state untouched |
| `v3.0.0 timeout`, or `start-at` | exit 1, no output | names what is missing |
| `timeout restart v3.0.0` | `--timeout restart` handed to helm | error |
| `timeout 0s v3.0.0` | expired deadline, instant failure | error |
| exported `HELM_TIMEOUT=bogus` | handed to helm unchecked | ignored |

## Notes

- A timeout implies `helm --wait`, so an `update` that supplies one now blocks on resources and fails within the timeout, rather than returning at once and falling through to the indefinite poll in `wait-for-combine`.
- A damaged `install-timeout` — unreadable, empty or not a usable duration — stops install and update runs, naming the file and saying to delete it or use `clean`. `uninstall` skips the timeout entirely, so recovery is never blocked.
- `restart` keeps a recorded timeout by design: it resets how far the install got, not saved settings.
- The value pass sees every token, so a value beginning with `v` is version-checked. Only `start-at` takes such a value, and no step name begins with `v`.
- `next-state` no longer writes `Done` into a missing state file — a latent path that left later runs skipping every step.

## Testing

`bash -n`, plus the script's own prologue driven against a throwaway `HOME`:

- rejected input leaves the state, the saved environment and any recorded timeout untouched
- record, reuse, override, `restart` keeps, `clean` drops, `Done` drops
- `restart`, `update`, `start-at`, `clean`, `uninstall`, prerelease versions, the dev flags and unknown options all behave as before

A full desktop install has not been re-run.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4351)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation flow handling so timeout settings are consistently applied during both fresh and resumed installations.
  * Ensured installation options are available before environment setup begins.

### DIFF
--- a/deploy/scripts/install-combine.sh
+++ b/deploy/scripts/install-combine.sh
@@ -315,9 +315,7 @@ if [ "${STATE}" != "Uninstall-combine" ] ; then
     if [ -f ${TIMEOUT_FILE} ] ; then
       ERROR_HINT="Delete ${TIMEOUT_FILE} or install with the clean option."
       HELM_TIMEOUT=`cat ${TIMEOUT_FILE}` || error "Cannot read ${TIMEOUT_FILE}."
-      if [ -n "${HELM_TIMEOUT}" ] ; then
-        check-opt-value timeout "${HELM_TIMEOUT}"
-      fi
+      check-opt-value timeout "${HELM_TIMEOUT}"
       ERROR_HINT=""
     fi
   else
--- a/deploy/scripts/install-combine.sh
+++ b/deploy/scripts/install-combine.sh
@@ -176,8 +176,9 @@ wait-for-combine () {
 next-state () {
   STATE=$1
   if [[ "${STATE}" == "Done" ]] ; then
-    # The recorded timeout applies to the install in progress, so it is
-    # discarded along with the state once that install is finished.
+    # A finished install has no more helm commands to run, so the recorded
+    # timeout is discarded along with the state. "restart" keeps it, since it
+    # resets how far the install got, while "clean" discards saved settings.
     rm -f ${STATE_FILE} ${TIMEOUT_FILE}
   else
     echo -n ${STATE} > ${STATE_FILE}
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm: Code-Rabbit seems to have caught an edge case, not sure if it's real.

> @jasonleenaylor reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4351#-:-P-LcTHd8VDZ4rHhFyXS:b-raqobv)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

PR #4352, opened 2026-08-19
Fix/laptop install first boot

imnasnainaec:
Devin: https://app.devin.ai/review/sillsdev/TheCombine/pull/4352

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4352)

### DIFF
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -3,6 +3,10 @@
 # completion here and only on success.  Doing it here rather than in the caller
 # means a manual run also counts, and stops the next container start from
 # redoing the whole import.
+#
+# Stop at the first failure and report it: The Combine cannot be used without the
+# semantic domains, so the database's postStart hook restarts the container on a
+# non-zero exit rather than leave a database that looks healthy without them.
 set -eo pipefail
 
 mongoimport -d CombineDatabase -c SemanticDomainTree /data/semantic-domains/tree.json --mode=merge --upsertFields=id,guid,lang
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -1,4 +1,7 @@
 #! /usr/bin/bash
+# The caller records the import as complete only if this script succeeds, so a
+# failed import must not be reported as success.
+set -eo pipefail
 
 mongoimport -d CombineDatabase -c SemanticDomainTree /data/semantic-domains/tree.json --mode=merge --upsertFields=id,guid,lang
 mongoimport -d CombineDatabase -c SemanticDomains /data/semantic-domains/nodes.json --mode=merge --upsertFields=id,guid,lang
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -1,6 +1,6 @@
 #! /usr/bin/bash
 # A partial import leaves the collections non-empty but incomplete, so record
-# completion here and only on success.  Doing it here rather than in the caller
+# completion here and only on success. Doing it here rather than in the caller
 # means a manual run also counts, and stops the next container start from
 # redoing the whole import.
 #
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -1,7 +1,11 @@
 #! /usr/bin/bash
-# The caller records the import as complete only if this script succeeds, so a
-# failed import must not be reported as success.
+# A partial import leaves the collections non-empty but incomplete, so record
+# completion here and only on success.  Doing it here rather than in the caller
+# means a manual run also counts, and stops the next container start from
+# redoing the whole import.
 set -eo pipefail
 
 mongoimport -d CombineDatabase -c SemanticDomainTree /data/semantic-domains/tree.json --mode=merge --upsertFields=id,guid,lang
 mongoimport -d CombineDatabase -c SemanticDomains /data/semantic-domains/nodes.json --mode=merge --upsertFields=id,guid,lang
+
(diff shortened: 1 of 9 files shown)


imnasnainaec:
## Splitting this PR: four stand-alone PRs, no conflicts

I built the split locally and tested it rather than eyeballing it: all 6 pairwise merges are clean, four-way
sequential merges are clean in every order I tried, and the union of the four reproduces this branch's tree exactly
(the only difference was a `git apply`-on-Windows file-mode artifact on the new script).

Everything is file-disjoint except two small overlaps, in `deploy/scripts/install-combine.sh` and
`installer/README.md`. Both were tested on their own: the regions are ~30 lines and one table row apart, and neither
conflicts.

### 1. Wait for cluster DNS before starting nginx

`nginx/init/05-wait-for-backend-dns.sh` (new), `nginx/templates/default.conf.template`,
`deploy/helm/thecombine/charts/frontend/templates/deployment-frontend.yaml`

From `c67b43e3` plus the nginx-template hunk of `fcb60fe4`. Shares no file and no behavior with the rest, and
`Dockerfile:66` already copies `nginx/init/*` into `/docker-entrypoint.d/`.

**QA & prod: affected.** Both changes live in the frontend image and the shared frontend chart, with no target
conditionals, so they ship to `qa-kube.thecombine.app` and `thecombine.app` as well. On a server whose cluster DNS is
already up, the script's first `getent` succeeds and it exits immediately, and the two-minute cap keeps it from ever
hanging a deploy. The new readiness probe changes when the frontend joins its Service endpoints on every target. All
three `proxy_pass` directives in the template currently point at `backend`, so the wait covers every cluster-internal
upstream as it stands, which is what the added comment is there to protect.

### 2. Database first boot: import record, postStart hook, readiness probe

`database/init/update-semantic-domains.sh`, `deploy/helm/thecombine/charts/database/templates/database.yaml`,
`docs/deploy/README.md`

From the database parts of `3360ecd4`, plus `e34efeb8`, `d5b578e7`, `75ff4485`, and parts of `fcb60fe4` / `e2fe421f`.

**QA & prod: affected, and this is the one to review with prod in mind.** The chart template has no target
conditionals and `updateStrategy` is `Recreate` on every profile, so on the first upgrade to this chart the database
pod is replaced and then held out of the `database` Service by the new readiness probe until its postStart hook
finishes. An existing QA or prod installation has no completion record, so that hook imports the semantic domains once
more: several minutes with no `database` endpoint. The documented manual rerun,
`kubectl -n thecombine exec deployment/database -- /opt/thecombine/update-semantic-domains.sh`, now also writes the
completion record, and the notes being edited in `docs/deploy/README.md` are the QA/prod deployment docs.
`setup_combine.py` as documented for QA/prod is invoked without `--wait`, so helm itself does not block on the import
there.

### 3. combinectl start/stop/status

`deploy/ansible/roles/support_tools/files/combinectl.sh`, the `Shutdown-combine` hunk of
`deploy/scripts/install-combine.sh`, the `update` row of `installer/README.md`

From the combinectl parts of `3360ecd4`, plus `dbda435d`, `dc93c0a2`, `64a94050`, `f6538023`, and parts of
`fcb60fe4` / `e2fe421f` / `55975511`.

**QA & prod: no effect on the server clusters.** The `support_tools` role that installs `combinectl` is imported only
by `playbook_desktop_setup.yml` (`hosts: localhost`) and `playbook_nuc_setup.yml` (`hosts: nuc`), never for the
`server` or `qa` groups. It does reach the NUCs, including the QA NUC and nuc1-3, so it is not purely a laptop change.

### 4. Installer waits

`deploy/scripts/install-combine.sh`, the `timeout` row of `installer/README.md`

From the wait rewrite in `3360ecd4`, plus `21e967ee`, `368848b0`, `5c53fc71`, `29ccf687`, `55975511`, and parts of
`fcb60fe4`.

**QA & prod: none.** `install-combine.sh` ships in `combine-installer.run`; QA and prod are deployed with
`setup_combine.py` per `docs/deploy/README.md`. The installer's own `server` option is the only server-shaped path it
touches.

### Ordering (functional, not textual)

1. **4 needs 2 first.** `wait-for-semantic-domains` polls `CombineDatabase.SemanticDomainImportStatus`, and nothing on
   master writes that record, so on its own 4 fails a first install after `IMPORT_CHECK_TIMEOUT_SECONDS` (120s) with
   the manual-import hint.
2. **2 changes what the installer's helm `--wait` covers.** The readiness probe keeps the database unready until the
   import finishes, so `install-the-combine` can exceed the default five-minute `--timeout`. The note documenting that
   is the `installer/README.md` `timeout` row, grouped above with 4; move it to 2 if 2 lands first. Tested: the two
   table rows split across separate PRs without conflict.
3. **The `combinectl stop || error` hunk has to travel with 3**, not 4. On master, `combine-stop` ends on
   `if systemctl is-active --quiet create_ap`, so it exits 1 whenever the hotspot is not running, and that caller
   landing alone would abort the installer at `Shutdown-combine`.

So 1 and 3 can go in any order at any time; 2 before 4.

### Cost of splitting

`3360ecd4` and `fcb60fe4` each straddle all four groups, so this cannot be done by cherry-picking commits. The four
branches have to be re-authored from the final state, which is how I built and tested them.

imnasnainaec:
Superseded by four stand-alone PRs, each verified to merge without conflict in any order:

- #4354 — wait for cluster DNS before starting nginx
- #4355 — redo the semantic domain import until it finishes
- #4353 — manage the deployments in `combinectl`
- #4356 — bound the installer's waits (stacked on #4355, which writes the record it waits on)

Split rationale and verification: https://github.com/sillsdev/TheCombine/pull/4352#issuecomment-5360053323

PR #4354, opened 2026-08-21
Wait for cluster DNS before starting nginx in the frontend

imnasnainaec:
**Reviewers can ignore all the stuff below.**

(Much of it is for context and details of this change within a larger effort.)

---

Split out of #4352, part 1 of 4 — see
https://github.com/sillsdev/TheCombine/pull/4352#issuecomment-5360053323 for the split and how it was verified.

nginx resolves the hostnames in its `proxy_pass` directives when it loads its configuration, and exits if any of them
cannot be resolved:

```text
[emerg] host not found in upstream "backend" in /etc/nginx/conf.d/default.conf:48
```

On a cold start the frontend container can be running before the cluster DNS is able to answer for the `backend`
service, which leaves the pod crash looping until the kubelet's restart backoff happens to line up with DNS being
ready.

- Wait for the name in a `docker-entrypoint.d` script, and start nginx anyway after two minutes, so that a real
  misconfiguration is still reported rather than leaving a container that runs but never serves.
- Add a readiness probe to the frontend, since with the wait in place a running container is no longer necessarily a
  serving one.
- Note at the `proxy_pass` directives that a new cluster-internal upstream has to be added to the wait. All three of
  them currently point at `backend`, so the wait covers every one as it stands.

### Scope

Three files, none of them touched by the other parts of #4352:

- `nginx/init/05-wait-for-backend-dns.sh` (new) — `Dockerfile` already copies `nginx/init/*` into
  `/docker-entrypoint.d/`, so nothing else is needed to run it
- `nginx/templates/default.conf.template`
- `deploy/helm/thecombine/charts/frontend/templates/deployment-frontend.yaml`

Independent of the other three parts in both directions, textually and functionally: it shares no file with them and
no behavior, so it can merge at any point in the sequence. Verified conflict-free against #4353 and the database part.

### QA and prod

Affected. Both changes are in the frontend image and the shared frontend chart, with no target conditionals, so they
ship to `qa-kube.thecombine.app` and `thecombine.app` as well as to the NUCs and desktop installs. On a server whose
cluster DNS is already up, the script's first `getent` succeeds and it exits immediately, and the two-minute cap keeps
it from ever hanging a deploy. The readiness probe changes when the frontend joins its Service endpoints on every
target.

### Testing

Not yet exercised on hardware. The failure it fixes needs a cold cluster start to reproduce, so it wants a laptop or
NUC install rather than a QA deploy, where DNS is up before the frontend rolls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4354)

## Bug Fixes

* Improved frontend deployment readiness checks to more reliably confirm that the service is available.
* Improved nginx startup handling when the backend hostname is not immediately resolvable, reducing premature connection failures.
* Added a startup wait period for backend DNS resolution, with a safe timeout so nginx can continue reporting unresolved upstream issues when necessary.

## Documentation

* Clarified hostname requirements for backend proxy configuration.

### DIFF
--- a/nginx/init/05-wait-for-backend-dns.sh
+++ b/nginx/init/05-wait-for-backend-dns.sh
@@ -18,15 +18,15 @@ if getent hosts "${BACKEND_HOST}" > /dev/null 2>&1 ; then
 fi
 
 echo "Waiting up to ${MAX_WAIT_SECONDS}s for '${BACKEND_HOST}' to resolve"
-WAITED=0
 until getent hosts "${BACKEND_HOST}" > /dev/null 2>&1 ; do
-    if [ "${WAITED}" -ge "${MAX_WAIT_SECONDS}" ] ; then
+    # SECONDS is wall clock since this shell started, so the cap covers the time
+    # getent spends blocking on resolver timeouts, not just the sleeps.
+    if [ "${SECONDS}" -ge "${MAX_WAIT_SECONDS}" ] ; then
         # Start nginx anyway so that it reports the problem itself, rather than
         # leaving a container that is running but never serving.
-        echo "'${BACKEND_HOST}' did not resolve after ${MAX_WAIT_SECONDS}s"
+        echo "'${BACKEND_HOST}' did not resolve after ${SECONDS}s"
         exit 0
     fi
     sleep 2
-    WAITED=$((WAITED + 2))
 done
-echo "'${BACKEND_HOST}' resolved after ${WAITED}s"
+echo "'${BACKEND_HOST}' resolved after ${SECONDS}s"
--- a/nginx/init/05-wait-for-backend-dns.sh
+++ b/nginx/init/05-wait-for-backend-dns.sh
@@ -1,0 +1,34 @@
+#! /bin/bash
+
+###############################################################
+# nginx resolves the hostnames in its proxy_pass directives when
+# it loads its configuration and exits if any of them cannot be
+# resolved:
+#
+#   [emerg] host not found in upstream "backend"
+#
+# On a cold start the frontend container can be running before
+# the cluster DNS is able to answer for the backend service, so
+# wait for the name before letting nginx start.
+###############################################################
+
+BACKEND_HOST=backend
+MAX_WAIT_SECONDS=120
+
+if getent hosts "${BACKEND_HOST}" > /dev/null 2>&1 ; then
+    exit 0
+fi
+
+echo "Waiting up to ${MAX_WAIT_SECONDS}s for '${BACKEND_HOST}' to resolve"
+WAITED=0
+until getent hosts "${BACKEND_HOST}" > /dev/null 2>&1 ; do
+    if [ "${WAITED}" -ge "${MAX_WAIT_SECONDS}" ] ; then
+        # Start nginx anyway so that it reports the problem itself, rather than
+        # leaving a container that is running but never serving.
+        echo "'${BACKEND_HOST}' did not resolve after ${MAX_WAIT_SECONDS}s"
+        exit 0
+    fi
+    sleep 2
+    WAITED=$((WAITED + 2))
+done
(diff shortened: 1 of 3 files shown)


imnasnainaec:
### Delivery note

Two of the three files here are image content rather than installer content:

- `nginx/init/05-wait-for-backend-dns.sh` and `nginx/templates/default.conf.template` are baked into the frontend
  image (`Dockerfile:63-66`)
- `deployment-frontend.yaml` is in the `deploy/` tree, so it ships in `combine-installer.run`

So the DNS wait itself only takes effect once a release publishes a new frontend image. An installer built from master
before then carries the readiness probe alongside an image with no wait script, because
`installer_release.yml` takes the image tag from `gh release view` while `actions/checkout` gives it master's `deploy/`
tree.

That pairing is harmless: the probe is a `tcpSocket` on port 80, so an old frontend serves or crash loops exactly as it
does today and the fix is simply absent. It does matter for verification, though — reproducing the crash loop and
confirming the fix needs a frontend image built from this branch, not the released tag.

imnasnainaec:
@coderabbitai review

nginx/init/05-wait-for-backend-dns.sh
```diff
@@ -0,0 +1,32 @@
...
+MAX_WAIT_SECONDS=120
+
+if getent hosts "${BACKEND_HOST}" > /dev/null 2>&1 ; then
+    exit 0
+fi
+
+echo "Waiting up to ${MAX_WAIT_SECONDS}s for '${BACKEND_HOST}' to resolve"
+WAITED=0
+until getent hosts "${BACKEND_HOST}" > /dev/null 2>&1 ; do
+    if [ "${WAITED}" -ge "${MAX_WAIT_SECONDS}" ] ; then
```
jasonleenaylor: `WAITED` counts the sleeps, not the time `getent` spends blocking -- so the cap is a floor.

It only bites in one of the two cold starts. If DNS is up and the Service just isn't there yet, NXDOMAIN returns fast, the loop ticks at ~2s and `WAITED` tracks wall clock. If DNS can't answer at all -- the case in your header comment -- each call blocks on resolver timeouts that `WAITED` never sees. So the bound holds in the case that doesn't need it.

A loop with a 3s-blocking probe and a nominal 10s cap ran 28s when I tried it. I couldn't measure real `getent` latency against your cluster.

`SECONDS` starts at zero in a new shell, so `[ "${SECONDS}" -ge "${MAX_WAIT_SECONDS}" ]` drops `WAITED` entirely, including `:30` and the echo at `:32`. Not `timeout 120 bash -c` -- it exits 124 and you'd have to rebuild the deliberate exit-0 fallback around it.

nginx/templates/default.conf.template
```diff
@@ -44,6 +44,7 @@ server {
     text/x-cross-domain-policy;
     # text/html is always compressed by gzip module
 
+    # Only "backend" is waited for in nginx/init/05-wait-for-backend-dns.sh.
```
jasonleenaylor: This says what is true now rather than what to do about it. The reader who needs the line is whoever adds a second upstream, and the thing they need is that the wait script needs editing too.

It's also above one of three -- `/export-hub` at `:61` and `/merge-hub` at `:70` proxy to the same host, and a new location near `:78` wouldn't see it.

```
# Adding a proxy_pass host? Add it to nginx/init/05-wait-for-backend-dns.sh
# too, or nginx will exit at startup when that name is not yet resolvable.
```

jasonleenaylor (approved):
Two nits inline. The mechanism is right and the readinessProbe is the right companion to it -- without that, a container mid-wait would take traffic.

_This review was assisted by Claude Opus 5._

jasonleenaylor (approved):
LGTM

_This review was assisted by Claude Opus 5._

PR #4355, opened 2026-08-21
Redo the semantic domain import until it finishes

imnasnainaec:
Split out of #4352, part 2 of 4 — see
https://github.com/sillsdev/TheCombine/pull/4352#issuecomment-5360053323 for the split and how it was verified.

The database pod's `postStart` hook imported the semantic domains whenever either collection was empty. A count cannot
tell a finished import from one killed part way through, so an interrupted import left the collections non-empty but
incomplete and was never redone.

- Guard the import with a completion record in `CombineDatabase.SemanticDomainImportStatus`, written only when the
  import finishes. The record is read by exit status, so anything short of a completed import imports again — a merge,
  and safe to repeat.
- Write the record from `update-semantic-domains.sh` rather than from the hook, so that the documented manual rerun
  counts too and does not leave the next pod start redoing the whole import.
- Fail the hook on a failed import, so the container restarts and retries. _The Combine_ cannot be used without the
  semantic domains, and a database that serves without them looks healthy.
- Add a readiness probe on a marker the hook writes last, so the pod stays out of the `database` Service until the
  replica set advertises the current pod IP — which changes on every restart, and which the backend needs since it
  connects with `?replicaSet=rs0`.
- Append to `/data/db/postStart.log` instead of truncating it, stamp each start, and trim to the last 200 lines, since
  the log is on the database's persistent volume with nothing rotating it.

### Scope

Four files, none of them touched by parts 1 or 4:

- `database/init/update-semantic-domains.sh`
- `deploy/helm/thecombine/charts/database/templates/database.yaml`
- `docs/deploy/README.md`
- the `timeout` row of `installer/README.md`

That last one is a deliberate move from where the split analysis put it. It documents that the installer's timeout has
to cover the semantic domain import, which is a consequence of the readiness probe here, so it belongs with the change
that causes it rather than with part 4. #4353 changes the `update` row of the same table; the two are verified
conflict-free, as is this branch against #4353 and part 1 as a whole.

### QA and prod

Affected, and this is the part to review with prod in mind. The chart template has no target conditionals and
`updateStrategy` is `Recreate` on every profile, so on the first upgrade to this chart the database pod is replaced and
then held out of the `database` Service by the new readiness probe until its `postStart` hook finishes. An existing QA
or prod installation has no completion record, so that hook imports the semantic domains once more: several minutes
with no `database` endpoint.

QA and prod deploys run `setup_combine.py` through `.github/actions/combine-deploy-update` without `--wait`, so helm
itself does not block on the import — but the endpoint gap is real and worth planning around. The manual rerun,
`kubectl -n thecombine exec deployment/database -- /opt/thecombine/update-semantic-domains.sh`, now also writes the
completion record.

### Testing

Not yet exercised on hardware. Worth covering both a fresh install, where the import runs inside the hook, and an
upgrade of an installation that already has the domains but no record, since that is the case QA and prod will hit
first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4355)

### DIFF
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -3,6 +3,10 @@
 # completion here and only on success.  Doing it here rather than in the caller
 # means a manual run also counts, and stops the next container start from
 # redoing the whole import.
+#
+# Stop at the first failure and report it: The Combine cannot be used without the
+# semantic domains, so the database's postStart hook restarts the container on a
+# non-zero exit rather than leave a database that looks healthy without them.
 set -eo pipefail
 
 mongoimport -d CombineDatabase -c SemanticDomainTree /data/semantic-domains/tree.json --mode=merge --upsertFields=id,guid,lang
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -1,4 +1,7 @@
 #! /usr/bin/bash
+# The caller records the import as complete only if this script succeeds, so a
+# failed import must not be reported as success.
+set -eo pipefail
 
 mongoimport -d CombineDatabase -c SemanticDomainTree /data/semantic-domains/tree.json --mode=merge --upsertFields=id,guid,lang
 mongoimport -d CombineDatabase -c SemanticDomains /data/semantic-domains/nodes.json --mode=merge --upsertFields=id,guid,lang
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -1,6 +1,6 @@
 #! /usr/bin/bash
 # A partial import leaves the collections non-empty but incomplete, so record
-# completion here and only on success.  Doing it here rather than in the caller
+# completion here and only on success. Doing it here rather than in the caller
 # means a manual run also counts, and stops the next container start from
 # redoing the whole import.
 #
--- a/database/init/update-semantic-domains.sh
+++ b/database/init/update-semantic-domains.sh
@@ -1,7 +1,11 @@
 #! /usr/bin/bash
-# The caller records the import as complete only if this script succeeds, so a
-# failed import must not be reported as success.
+# A partial import leaves the collections non-empty but incomplete, so record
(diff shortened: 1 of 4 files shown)


imnasnainaec:
### Delivery note — publish the release before building an installer

`database/init/update-semantic-domains.sh` is baked into the database image (`database/Dockerfile`, into
`/opt/thecombine/`), while `database.yaml` ships in the `deploy/` tree inside `combine-installer.run`. This is the PR
where those two stop being independent of each other.

`installer_release.yml` takes the image tag from `gh release view` — the latest published release — while
`actions/checkout` gives it master's `deploy/` tree. Those normally coincide, since the installer build follows a
release deploy. But an installer built from master after this merges and before the next release pairs the new chart
with the previous release's images.

In that pairing this breaks hard rather than degrading. `e34efeb8` moved the completion-record write out of the
`postStart` hook and into the script, so a new hook calling an old script records nothing, and every pod start reimports
forever. With #4356 also in, `wait-for-semantic-domains` never passes: the install fails after
`IMPORT_CHECK_TIMEOUT_SECONDS` with a manual-import hint that cannot help either, because the manual run is the same old
script.

`combinectl update <tag>` reaches the same state from the other direction, since it moves image tags and leaves the
charts alone.

Two ways out:

1. Publish the release before building any installer that carries this chart.
2. Restore the hook-side record write as a fallback, after the script succeeds — the line `e34efeb8` removed:

   ```sh
   mongosh --quiet --host 127.0.0.1 --eval "db.getSiblingDB('CombineDatabase').SemanticDomainImportStatus.replaceOne({ _id: 'semantic-domains' }, { _id: 'semantic-domains', completed: true }, { upsert: true });"
   ```

   Keeping both writes is idempotent, and each covers a case the other does not: the script's write is what makes a
   manual rerun count, the hook's is what makes an older image work.

Option 2 is a one-line change to this PR and removes the ordering constraint entirely, which is the way I would lean.

---

Update: Option 2 applied.

PR #4365, opened 2026-09-14
Fix merge of a new word with no id

imnasnainaec:
## Problem

Reported by a user reviewing deferred duplicates:

> as I am reviewing the deferred entries, I'm getting this error.
> 400 https://thecombine.app/v1/projects/.../merge

When a sense is dragged into the empty new-word column, the merge tree key is a
`crypto.randomUUID()`, so `state.data.words[wordId]` misses and the parent is built from
`newWord()` — which has `id: ""`. `MergePrepParent` leaves it empty, since no child guid
matches.

Since #4198, `ReplaceFrontierWithSession` sends every parent through
`UpdateFrontierWithSession(..., createIfNotFound: true, ...)`, which unconditionally
filters the Frontier on that id. `Word.Id` is `[BsonRepresentation(BsonType.ObjectId)]`,
so the driver throws while rendering the filter:

```
System.FormatException : '' is not a valid 24 digit hex string.
```

The pre-#4198 `MergeService.Merge` guarded the lookup with
`if (childrenIds.Contains(parentId))`, so an empty parent id fell straight through to
`Create`. The transaction rewrite dropped that guard.

`MergeController.MergeWords` catches every exception and returns `BadRequest("Merge
failed.")` without logging, so there is no server-side trace. The frontend's axios
interceptor renders the snackbar the user quoted, and `mergeAll()` has no catch, so
`SaveDeferButtons.saveContinue` never reaches `next()` — the button keeps spinning and the
user is stuck on that set.

This affects both Merge Duplicates and Review Deferred Duplicates; the reporter happened to
hit it on the latter.

## Fix

Skip the Frontier lookup when the word has no id. There is no predecessor to replace, so
the word is created — which is what the caller wants in both directions:

- `createIfNotFound: true` (merge): the new parent is created.
- `createIfNotFound: false` (`WordService.Update`): returns `null`, so `WordController`
  answers 404 instead of throwing.

## Testing

Two integration tests in `WordRepositoryTests`, both of which fail on `master` with
`System.FormatException : '' is not a valid 24 digit hex string.`:

- `TestReplaceFrontierCreatesWordWithNoId`
- `TestUpdateFrontierByWordWithNoIdReturnsNull`

Full `WordRepositoryTests` fixture: 61 passed.

## Not addressed here

Two related findings from the same investigation, left for separate PRs:

1. **Stale Frontier ids now hard-fail.** `ReplaceFrontierWithSession` throws `ArgumentException("All old words being replaced must exist in the Frontier")` if a merge child has left the Frontier. Pre-#4198 this was tolerated — the `DeleteFrontierWord` result was ignored. Review Deferred Duplicates is most exposed, because all steps load up front and are never refreshed: graylist entries may overlap (`AddToMergeGraylist` only removes entries fully *contained* in the new set), so merging one step invalidates a later one. The mirror case is worse than a 400 — when the *parent* carries the stale id, `createIfNotFound: true` silently creates a duplicate word.
2. **`MergeWords` swallows the exception.** No log, no OTel error, no reason on the response, which is why this report could not be diagnosed from production data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4365)

## Summary by CodeRabbit

- **Bug Fixes**
  - Frontier updates with words missing an ID no longer trigger database format errors.
  - Updating a frontier with a missing ID now safely returns no result without saving invalid data.
  - Replacing a frontier word without an ID now assigns a valid ID and preserves frontier counts.

- **Tests**
  - Added coverage for frontier update and replacement behavior involving words without IDs.


jasonleenaylor (approved):
LGTM

PR #4371, opened 2026-09-16
Wait for the backend before restoring from a backup

imnasnainaec:
A restore on QA died at "Copy the backend files" with a raw `CalledProcessError`:
`kubectl` had no backend pod in phase `Running`, and `{.items[0].metadata.name}`
errors on an empty list. By then step 4 had already dropped and replaced the
database, leaving QA half-restored.

QA redeploys on every push to `master` and inherits `updateStrategy: Recreate`
with `imagePullPolicy: Always`, so its backend pod has a window with nothing
Running. Fetching and unpacking a 7.5 GB backup is long enough to overlap one.

## Changes

`combine_restore.py`

- Wait for the database and backend deployments up front, as `combine_backup.py`
  already does, so an unavailable cluster fails before the download.
- Wait again and resolve both pod ids together before the restore drops
  anything. A rollout landing mid-restore is now waited out rather than failing
  after the destructive step.

`combine_app.py`

- `get_pod_id` prints `{.items[*].metadata.name}` and indexes in Python,
  returning `""` on no match instead of dying on kubectl's out-of-bounds error.
  A miss is not cached, so a retry re-queries. This also makes the existing
  `Cannot find the ... container` guards in both scripts reachable.

## Testing

- `black`, `isort`, `flake8`, and `mypy --platform linux maintenance/scripts` clean.
- `get_pod_id` exercised against stubbed kubectl output: zero/one/two pods, the
  `instance` selector, and re-query after a miss.
- Not run against a live cluster.

## Not addressed

- The wait budget is `wait_time` (120s). A cold backend image pull under
  `Recreate` can exceed it; the restore then exits with a real message instead of
  a traceback, but still fails. Happy to bump `waitTime` if that proves tight.
- Node ephemeral storage. The maintenance pod declares no `ephemeral-storage`
  request and the restore copies the dump into the database container's root
  filesystem rather than its PVC, so eviction under DiskPressure is a plausible
  second cause of a missing pod. Separate change; see also
  `copilot/reduce-ephemeral-storage-backup-job` for the backup half.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4371)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restore reliability when services are temporarily unavailable or restart during the process.
  * Missing service instances are now handled gracefully instead of causing the maintenance operation to stop unexpectedly.
  * Service availability is rechecked before restoring the database and copying backend files, reducing failures caused by deployment rollouts.
  * Restore failures now provide clearer status information, including when an operation may be partially complete.


maintenance/scripts/combine_restore.py
```diff
@@ -73,6 +83,10 @@ def main() -> None:
     combine = CombineApp()
     aws = AwsBackup(bucket=aws_bucket)
     step = ScriptStep()
+    wait_time = int(os.getenv("wait_time", 60))
```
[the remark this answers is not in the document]
imnasnainaec: The chart already sets `wait_time` to 120 (`maintenance/values.yaml`), so this fallback never applies in the maintenance pod — and it matches `combine_backup.py:47`. The 120s note in the description flagged that budget as possibly too short, not as a requirement for this literal.

🤖 Written by Claude Opus 5 (1M context) via [Claude Code](https://claude.com/claude-code)

jasonleenaylor (commented):
> @jasonleenaylor reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4371)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

maintenance/scripts/combine_restore.py
```diff
@@ -177,7 +197,10 @@ def safe_extract(
         logging.debug(f"stdout:\n{rm_proc.stdout.strip()}")
 
         step.print("Copy the backend files.")
-        backend_pod = combine.get_pod_id(CombineApp.Component.Backend)
+        # The database is already replaced, so failing here leaves the restore half done:
+        # wait out any rollout rather than fail on a pod id that predates it.
+        wait_for_combine(wait_time)
```
jasonleenaylor: This call is past the point of no return -- the database was dropped and replaced at line 181 -- but it exits through the same `wait_for_combine` as the two calls at :89 and :171, so the operator gets "The database or the backend is not available" with nothing to say the restore is half done. That is the state this PR is fixing.

A second argument on the helper, or a distinct message here, would cover it.

jasonleenaylor (approved):
One question inline about the last wait.

_This review was assisted by Claude Opus 5._

maintenance/scripts/combine_restore.py
imnasnainaec: Done.

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/TheCombine/4371)* status: 1 of 2 files reviewed, all discussions resolved (waiting on jasonleenaylor).

